PR #1, opened 2021-02-16
Add Cloudformation file for deploying state machine with simple pipeline

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-507

N/A

N/A

N/A

# Changes
### Code changes
A CloudFormation file can deploy a state machine that can call into the Kubernetes cluster.

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

- [x] Relevant wiki pages created/updated

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/cloudformation-states.yaml
+++ b/cloudformation-states.yaml
@@ -1,0 +1,110 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: Set up an executable pipeline state machine
+
+Parameters:
+  Environment:
+    Type: String
+    Default: staging
+    AllowedValues:
+      - staging
+      - production
+    Description: The environment for which the cluster need to be created.
+  ExperimentId:
+    Default: '6379'
+    Type: String
+    Description: The ID of the experiment the system is created for.
+
+Resources:
+  PipelineStateMachine:
+    Type: "AWS::StepFunctions::StateMachine"
+    Properties:
+      DefinitionString:
+        !Sub
+          - |-
+            {
+              "Comment": "N/A",
+              "StartAt": "DeleteCompletedJobs",
+              "States": {
+                "DeleteCompletedJobs": {
+                  "Type": "Task",
+                  "Comment": "Deletes all the preivous server jobs that are already completed.",
+                  "Resource": "arn:aws:states:::eks:call",
+                  "Parameters": {
+                    "ClusterName": "biomage-${Environment}",
+                    "CertificateAuthority.$": "$.certificateAuthority",
+                    "Endpoint": "${Endpoint}",
+                    "Method": "DELETE",
+                    "Path": "/apis/batch/v1/namespaces/default/jobs",
+                    "QueryParameters": {
+                      "fieldSelector": [
+                        "status.successful=1"
+                      ]
+                    }
+                  },
+                  "Next": "LaunchKubernetesJobIfNotExists"
+                },
+                "LaunchKubernetesJobIfNotExists": {
+                  "Type": "Task",
+                  "Comment": "Attempts to create a Kubernetes Job for the pipeline server. Will swallow a 409 (already exists) error.",
+                  "Resource": "arn:aws:states:::eks:call",
+                  "Parameters": {
+                    "ClusterName": "biomage-${Environment}",
+                    "CertificateAuthority.$": "$.certificateAuthority",
+                    "Endpoint": "${Endpoint}",
+                    "Method": "POST",
+                    "Path": "/apis/batch/v1/namespaces/default/jobs",
+                    "RequestBody": {
+                      "apiVersion": "batch/v1",
+                      "kind": "Job",
+                      "metadata": {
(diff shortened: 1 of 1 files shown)


cloudformation-states.yaml
```diff
@@ -0,0 +1,110 @@
...
+              "Comment": "N/A",
+              "StartAt": "DeleteCompletedJobs",
+              "States": {
+                "DeleteCompletedJobs": {
+                  "Type": "Task",
+                  "Comment": "Deletes all the preivous server jobs that are already completed.",
+                  "Resource": "arn:aws:states:::eks:call",
+                  "Parameters": {
+                    "ClusterName": "biomage-${Environment}",
+                    "CertificateAuthority.$": "$.certificateAuthority",
```
marcellp, line 34: `certificateAuthority` is an input parameter that must be added when the pipeline is run. This is because there is no easy way to get the CA from the CloudFormation stack outputs for the cluster. Hopefully we can resolve this problem from the API in the next ticket so we don't need to do this.

cloudformation-states.yaml
```diff
@@ -0,0 +1,110 @@
...
+                      "spec": {
+                        "template": {
+                          "metadata": {
+                            "name": "remoter-server-${ExperimentId}"
+                          },
+                          "spec": {
+                            "containers": [
+                              {
+                                "name": "remoter-server",
+                                "image": "242905224710.dkr.ecr.eu-west-1.amazonaws.com/pipeline:refs-heads-master-236bab58afda4a763bc6a72514247e228fae5c75-remoter-server"
```
marcellp, line 71: This is obviously hardcoded at this point, will be changed in the API.

cloudformation-states.yaml
```diff
@@ -0,0 +1,110 @@
...
+                }
+              }
+            }
+          - Endpoint:
+              Fn::ImportValue:
+                !Sub "eksctl-biomage-${Environment}-cluster::Endpoint"
+              
+      RoleArn:
+        Fn::ImportValue:
+          !Sub "biomage-state-machine-role-${Environment}::ARN"
```
marcellp, line 110: This is an ARN role that is created in `iac`. This will probably need to be changed for the production version as well, but this way it's a standalone CF file that can work.

marcellp (commented):
Some context for the cloudformation file.

cloudformation-states.yaml
```diff
@@ -0,0 +1,110 @@
...
+Resources:
+  PipelineStateMachine:
+    Type: "AWS::StepFunctions::StateMachine"
+    Properties:
+      DefinitionString:
+        !Sub
+          - |-
+            {
+              "Comment": "N/A",
+              "StartAt": "DeleteCompletedJobs",
```
cosa65, line 26: mainly to learn a little bit more about this: why does the pipeline start by deleting previous jobs? couldn't this be managed somewhere else?

cloudformation-states.yaml
```diff
@@ -0,0 +1,110 @@
...
+                      "ErrorEquals": [ "EKS.409" ],
+                      "ResultPath": "$.error-info",
+                      "Next": "Wait"
+                    }
+                  ],
+                  "Next": "Wait"
+                },
+                "Wait": {
+                  "Type": "Wait",
+                  "Seconds": 5,
```
cosa65, line 99: the ticket mentioned waiting for 10 seconds, does this difference matter?

cloudformation-states.yaml
marcellp: It is not possible, not easily at least.

What this does is it deletes previous /completed/ jobs. Unfortunately the way Kubernetes structures Jobs is that their output persists even after they fail or succeed. This makes sense for most applications where you want to review the results after the fact.

Unfortunately for our purposes this does not work, because we want to have a Job running all the time.

Imagine if we did not delete completed jobs, and the first step is to create a Job. If the creation throws a 409 (already exists), then catch the error and fail silently, knowing that we have a job that exists and can send messages to. The problem with this is that the 409 will also be thrown if there is already a *completed* job. Therefore, there will be no Pods running and the future steps will fail.

Therefore, a necessary step for this to work is before the pipeline runs, we clean up all completed Jobs that match our specifications. This way, when the next steps throws a 409, we know that there is an *active, running* job that we can send stuff to.

There is obviously a race condition here (what if the Job shuts down in between the two steps). There is not much we can do about this and it's not the end of the world -- if no jobs are deleted in step 1, then the job finished, then in step 2 we find an already existing job and falsely assume it is running, we will fail on the first actual step because of a connection error and the state machine will fail, but the incidence of this will be quite low.

cloudformation-states.yaml
marcellp: It does not really matter, it was just a way to add more than one step to the pipeline.

cosa65 (approved):

PR #28, opened 2021-04-02
Make pipeline use Step Function Activities

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-709
https://biomage.atlassian.net/browse/BIOMAGE-764

https://github.com/biomage-ltd/api/pull/83

TBA

N/A

# Changes
### Code changes
Pipeline repository is substantially refactored. It is now using only one project and it executes tasks based on Step Functions activities.

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

- [x] Relevant wiki pages created/updated

### Approvers

### Just before merging:

### Optional


pipeline.code-workspace
```diff
@@ -1,17 +0,0 @@
-{
-	"folders": [
-		{
-			"path": "."
-		}
-	],
```
ivababukova, line 6: why is this getting removed?

qc-runner/src/init.r
```diff
@@ -214,4 +217,37 @@ wrapper <- function(input_json) {
...
+                states$send_task_failure(
+                    taskToken = taskToken,
+                    error = 'R error',
+                    cause = e
+                )
+            }
+        )
+    }
+
+    message(result)
```
xverges, line 250: Does not look like this is defined

qc-runner/src/init.r
```diff
@@ -214,4 +217,37 @@ wrapper <- function(input_json) {
     return(message_id)
 }
 
-message("New wrapper loaded.")
+init <- function() {
+    pipeline_config <- load_config('host.docker.internal')
+    states <- paws::sfn(config=pipeline_config$aws_config)
+
+    repeat {
```
xverges, line 224: What's our plan to cleanup pipeline servers?

local-runner/cf-local-container-launcher.yaml
```diff
@@ -28,26 +28,23 @@ Resources:
...
-              task = f'\"{task}\"'
-
             docker_command = ' '.join([
               "docker run --rm",
-              "__DOCKER_GATEWAY_HOST__",
               f"--name {event['name']}-{random_string(10)}", 
-              f"{'-d -p 6969:6969' if event['detached'] else ''}", 
-              f"--env SAMPLE_ID={event.get('sampleUuid', '')}",
-              event['image'],
-              task
```
xverges, line 43: So many simplifications evrywhere! 🤩
You'll also be able to do some cleanup in `app.py`

marcellp: just did so, thanks for the heads up!

qc-runner/src/init.r
marcellp: Right, deleting it just now.

qc-runner/src/init.r
marcellp: The plan is to simply quit the server if there was no message received by it. GetActivityTask [polls for a minute](https://docs.aws.amazon.com/step-functions/latest/apireference/API_GetActivityTask.html) for new work, and if it can't find anything it will return an [empty string](https://paws-r.github.io/docs/sfn/get_activity_task.html) in R in particular.

There is a way to do more advanced error handling and recovery, but because we are not persisting intermediate count matrices to S3, it is probably better to just leave it as is and just quit successfully if we get an empty task token. Mind you, right now with the `tryCatch` block if there is an error being thrown in the step it gets automatically written off as a failure as we catch exceptions and immediately return a failed activity. If there's a condition where we'd end up with no task getting to the worker in a minute it probably makes sense to just nuke the whole thing anyways.

pipeline.code-workspace
marcellp: Because we are not using it anymore. Each project contained in pipeline (the worker and the local runner) has its own workspace file now.

qc-runner/src/init.r
xverges: Makes sense, thanks. I tend to forget that every pipeline execution starts everything from scratch: new state machine, new state machine execution, new runner job...

BTW: _new runner job_ makes me think that, after remoter client/server setup has been gone, we don't need to have a service as a part the HelmRelease

marcellp: That would be correct, yes. I'll remove it from the chart just now.

xverges (approved):

PR #38, opened 2021-04-13
Downsampling plots

JuanluOnieva-biomage:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-788?atlOrigin=eyJpIjoiMjM5ODc1NDk2ZTQ4NDk2NWFkZTQxMWVkNmQxZDQxYmIiLCJwIjoiaiJ9

We need to downsample the plotData to avoid the UI being slow and laggy. The current PR include the capability to downsample it. As a proposal, I have added a parameter to each step function related to the number of cells to downsample. @aerlaut will take care of tuning this parameter. 

[Commit 6a0c8d6]: Agi asked me to compute the transformed value of the UMIs for the knee plot in the pipeline side. Hence, a PR in the UI is required by adapting the new plotData. Now, the input of the knee plot is `log_u` (instead of `u`) and `rank`

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Tested locally with Inframock
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/qc-runner/src/cellSizeDistribution.r
+++ b/qc-runner/src/cellSizeDistribution.r
@@ -82,11 +82,8 @@ task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsamp
     # 3483 6019 3892 3729 4734 3244
     
     plot2_data <- seq_along(plot1_data)
-    plot2_data <- unname(map2(plot1_data,plot2_data,function(x,y){c("u"=x,"rank"=y)}))
-    # Once the UI is updated and takes into account that the transformation is doing in the R side we can uncomment
-    # the following line and remove the previous one.
-    # plot1_data_transformed <- (log(plot1_data)/log(10))*2200
-    # plot2_data <- unname(map2(plot1_data_transformed,plot2_data,function(x,y){c("log_u"=x,"rank"=y)}))
+    plot1_data_transformed <- (log(plot1_data)/log(10))*2200
+    plot2_data <- unname(map2(plot1_data_transformed,plot2_data,function(x,y){c("log_u"=x,"rank"=y)}))
     plot1_data <- lapply(unname(plot1_data),function(x) {c("u"=x)})
 
     # Downsample plotData
--- a/qc-runner/src/cellSizeDistribution.r
+++ b/qc-runner/src/cellSizeDistribution.r
@@ -83,6 +83,10 @@ task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsamp
     
     plot2_data <- seq_along(plot1_data)
     plot2_data <- unname(map2(plot1_data,plot2_data,function(x,y){c("u"=x,"rank"=y)}))
+    # Once the UI is updated and takes into account that the transformation is doing in the R side we can uncomment
+    # the following line and remove the previous one.
+    # plot1_data_transformed <- (log(plot1_data)/log(10))*2200
+    # plot2_data <- unname(map2(plot1_data_transformed,plot2_data,function(x,y){c("log_u"=x,"rank"=y)}))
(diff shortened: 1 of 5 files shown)


qc-runner/src/cellSizeDistribution.r
```diff
@@ -82,9 +82,20 @@ task <- function(seurat_obj, config, task_name, sample_id) {
     # 3483 6019 3892 3729 4734 3244
     
     plot2_data <- seq_along(plot1_data)
-    plot2_data <- unname(map2(plot1_data,plot2_data,function(x,y){c("u"=x,"rank"=y)}))
+    plot1_data_transformed <- (log(plot1_data)/log(10))*2200
```
seb-mueller, line 85: I suppose this is a design choice to have the log computed on the pipeline side rather than the UI side (i.e. through a checkbox to activate/dactivate log transformation?

qc-runner/src/cellSizeDistribution.r
```diff
@@ -50,7 +50,7 @@ generate_default_values_cellSizeDistribution <- function(seurat_obj, config) {
   return(sample_subset$nCount_RNA)
 }
 
-task <- function(seurat_obj, config, task_name, sample_id) {
+task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsample = 5000) {
```
seb-mueller, line 53: 5000 is the default, but if other values are desired, how are they controlled? Via DynamoDB?

seb-mueller (approved):
Looks good, is there a staging link to test?

qc-runner/src/cellSizeDistribution.r
JuanluOnieva-biomage: Yes, was a design from the engineer side. I think that it is something that we are not going to change, because for the knee plot we use always the log transformed. But, I am happy to control it by a parameter if you think that we need to keep also the non transformed list

qc-runner/src/cellSizeDistribution.r
JuanluOnieva-biomage: Currently, these values are going to be tested by @aerlaut, but for now they are HARDCODED. For a long-term we need to think how we control them. Maybe we can add an entry in the config

qc-runner/src/cellSizeDistribution.r
seb-mueller: Ok, I think we can park this discussion for another time, it's ok for now. My thinking was to also be able to control the y axis with a log checkbox going forward. Let's discuss in the bionfo meeting.

qc-runner/src/cellSizeDistribution.r
seb-mueller: Ok, just wanted to know the state, fine for me for the time being.

qc-runner/src/cellSizeDistribution.r
ogibson-biomage: Juanlu, could you explain a bit more about that normalization? why the 2200 hardcoded value?

JuanluOnieva-biomage: Agi gave me this formula, I think that It should be explained in the design document or probably Seb has more knowledge here since he developed this filter.

aerlaut: The formula was the in the UI, I also have no idea what the 2200 is for,

qc-runner/src/cellSizeDistribution.r
aerlaut: I've checked and the uploaded file plot data file contains 5000 samples. The UI is still slow though,

ivababukova:
Closing that one, because we currently have #41 that builds on top of this one

PR #48, opened 2021-04-27
Fix barcode ranks plot and downsampling

alexvpickering:
# Background
#### Link to issue 

https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&modal=detail&selectedIssue=BIOMAGE-865

https://ui-alex-pipeline48.scp-staging.biomage.net/experiments/52df58315903e78c8350f1abbd27e358/data-processing

https://github.com/biomage-ltd/data-ingest/pull/22

In addition to fixing the issue in the above ticket, this also fixes an unrealized bug in the classifier plot. Prior to this, the plot gave you the impression that the majority of droplets have low FDR values:

image

After these changes it correctly reflects that the majority of droplets are empty (FDR density highest near 1):

image

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/qc-runner/src/cellSizeDistribution.r
+++ b/qc-runner/src/cellSizeDistribution.r
@@ -50,7 +50,7 @@ generate_default_values_cellSizeDistribution <- function(seurat_obj, config) {
   return(sample_subset$nCount_RNA)
 }
 
-task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsample = 6000, percent_downsample = 20) {
+task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsample = 6000) {
     print(paste("Running",task_name,sep=" "))
     print("Config:")
     print(config)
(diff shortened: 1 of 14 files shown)


qc-runner/src/cellSizeDistribution.r
```diff
@@ -74,29 +74,30 @@ task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsamp
...
+
+    # umi histogram plot
+    numis <- sort(sample_subset$nCount_RNA, decreasing = TRUE)
+    plot1_data <- lapply(unname(numis),function(x) {c("u"=x)})
+
+    # barcode ranks plot data
+    # unique average ranks maintain plot shape in case of downsampling
+    ranks <- rank(-numis, ties.method = 'average')
+    dups <- duplicated(ranks)
+    ranks <- ranks[!dups]
```
seb-mueller, line 86: Whats the rational to remove duplicated ranks? Is it for cosmetic reasons? I'm just concerend that this would bias the plot, i.e. in low abundant cells all with say a count of 11 would be condensed to only 1 cell.

qc-runner/src/classifier.r
```diff
@@ -100,17 +99,20 @@ task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsamp
       print("Info: empty-drops table of FDR threshold categories (# UMIs for a given threshold interval")
       print(table(obj_metadata$orig.ident, cut(obj_metadata$emptyDrops_FDR, breaks = c(-Inf,0,0.0001,0.01,0.1,0.5,1,Inf)), useNA="ifany"))
       print("How many barcodes should be filtered out for this sample (#FALSE):")
-      print(table(sample_subset$emptyDrops_FDR <= FDR))
+      
+      # prevents filtering of NA FDRs if FDR=1
+      ed_fdr <- sample_subset$emptyDrops_FDR
+      ed_fdr[is.na(ed_fdr)] <- 1
```
seb-mueller, line 105: Good catch, emptyDrops is assigning NA for low abudant cells, making it explicit to 1 is way better for our purposes :)

qc-runner/src/classifier.r
```diff
@@ -100,17 +99,20 @@ task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsamp
...
+      print(table(ed_fdr <= FDR))
```
seb-mueller, line 106:
```suggestion
      print(paste0("Number of droplets with an emptyDrops-FDR <= ", FDR, " :"))
      print(table(ed_fdr <= FDR, useNA="ifany"))
```

qc-runner/src/numGenesVsNumUmis.r
```diff
@@ -33,7 +33,8 @@ source('utils.r')
 #'                                - p.level: which refers to  confidence level for deviation from the main trend
 #' @export return a list with the filtered seurat object by numGenesVsNumUmis, the config and the plot values
 
-task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsample = 5000, percent_downsample = 20){
+
+task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsample = 5000){
```
seb-mueller, line 37:
```suggestion
task <- function(seurat_obj, config, task_name, sample_id, num_cells_to_downsample = 6000){
```
Shoudn't this be also 6000 as the others?

seb-mueller (commented):
Looks good in general, and the staging link seems to do the right thing.
A final test would be on an unfiltered multi-sample dataset such as `352cb8d159d650c13bc4acd5f7442993`
I've loaded it into staging, it seems to be fine in general, but the data cloud (FDR=1) seems missing from the plot (see below). Maybe it's because of ingesting has filtered them all out in which case it's fine, just wanted you to have a quick look:
 
https://ui-alex-pipeline48.scp-staging.biomage.net/experiments/352cb8d159d650c13bc4acd5f7442993/data-processing

image

qc-runner/src/cellSizeDistribution.r
alexvpickering: The rationale is that e.g. the we could have ~600k drops with the majority being <10 UMIs. These droplets have the same rank and the same UMI so they have the exact same coordinates on the kneeplot. By removing them we can get down to ~8k cells and then then subsequent downsampling to 6k will still yield a kneeplot with the exact shape instead of being heaving skewed towards high-rank low UMI droplets (and dropping the high UMI low rank cells we are interested in keeping so that they contribute to the shape). We only do this for the knee plot.

qc-runner/src/classifier.r
alexvpickering: This info is already printed above a few lines. I'll bring it down so that it is more obvious

qc-runner/src/numGenesVsNumUmis.r
alexvpickering: Good catch. Not sure why this one is set to 5000. Maybe @JuanluOnieva-biomage or @aerlaut knows?

alexvpickering:
> Looks good in general, and the staging link seems to do the right thing.
> A final test would be on an unfiltered multi-sample dataset such as `352cb8d159d650c13bc4acd5f7442993`
> I've loaded it into staging, it seems to be fine in general, but the data cloud (FDR=1) seems missing from the plot (see below). Maybe it's because of ingesting has filtered them all out in which case it's fine, just wanted you to have a quick look:
> 
> https://ui-alex-pipeline48.scp-staging.biomage.net/experiments/352cb8d159d650c13bc4acd5f7442993/data-processing
> 
> image

Yeah I just downloaded this one. The minimum `nCount_RNA` is 219 so it needs to be re-ingested. Also `sum(is.na(blah$emptyDrops_FDR)) = 0`

qc-runner/src/cellSizeDistribution.r
ogibson-biomage: Are we erasing the repeated ranks after or before we calculate the inflection points? This should be fine if we are doing it post calculation of the inflection points. Also, we need to make sure that this change to the ranks doesn't negatively impact the positioning of the vertical threshold line in the plot bc it's inferred from the min cell size.

alexvpickering: We are erasing the repeated rank only for the knee plot data. The calculation of the inflection point is unchanged

ogibson-biomage: Great! Thanks for clarifying alex, I'll make sure we aren't unintentionally breaking the UI threshold line and take a look at the staging links and if all looks good I'll approve and merge.

qc-runner/src/numGenesVsNumUmis.r
ogibson-biomage: Certainly looks like a typo

alexvpickering: I changed it to `6000` unless we hear otherwise

seb-mueller (approved):
All good from my side, nice work Alex!

PR #49, opened 2021-04-27
fix makefile for linux

alexvpickering:
Makefile needed some [black magic](https://lists.freebsd.org/pipermail/freebsd-questions/2012-September/244810.html) to get `awk` command to work


xverges (approved):
That `$$` thing was nasty!

PR #51, opened 2021-04-27
update debug stuff

alexvpickering:
Changes:

- saves objects using parameter names instead of argument names (makes more sense as can directly use to step through function)
- reflect Makefile usage for debug in README

### DIFF
--- a/qc-runner/src/init.r
+++ b/qc-runner/src/init.r
@@ -105,13 +105,16 @@ handle_debug <- function(scdata, config, task_name, sample_id, debug_config) {
     is_debug <- debug_config$step %in% c(task_name, 'all')
     
     if (is_debug) {
+        # variable names used by functions
+        seurat_obj <- scdata
+        num_cells_to_downsample <- 6000
+
         sample_str <- ifelse(sample_id == '', '', paste0('_', sample_id))
         fname <- paste0(task_name, sample_str, '.RData')
         fpath_cont <- file.path('/debug', fname)
         fpath_host <- file.path(debug_config$path, fname)
-        
         message(sprintf('⚠️ DEBUG_STEP = %s. Saving arguments.', task_name))
-        save(scdata, config, task_name, sample_id, file = fpath_cont)
+        save(seurat_obj, config, task_name, sample_id, num_cells_to_downsample, file = fpath_cont)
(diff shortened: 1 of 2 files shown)


qc-runner/src/init.r
```diff
@@ -105,13 +105,16 @@ handle_debug <- function(scdata, config, task_name, sample_id, debug_config) {
     is_debug <- debug_config$step %in% c(task_name, 'all')
     
     if (is_debug) {
+        # variable names used by functions
+        seurat_obj <- scdata
+        num_cells_to_downsample <- 6000
```
xverges, line 110: Not strictly related to this PR, but I found https://github.com/biomage-ltd/pipeline/blob/master/qc-runner/src/numGenesVsNumUmis.r#L36 using 5000. Maybe rather than having a default parameter the downsample number should be set from `init.r`?

qc-runner/src/init.r
```diff
@@ -105,13 +105,16 @@ handle_debug <- function(scdata, config, task_name, sample_id, debug_config) {
...
+
         sample_str <- ifelse(sample_id == '', '', paste0('_', sample_id))
         fname <- paste0(task_name, sample_str, '.RData')
         fpath_cont <- file.path('/debug', fname)
         fpath_host <- file.path(debug_config$path, fname)
-        
         message(sprintf('⚠️ DEBUG_STEP = %s. Saving arguments.', task_name))
-        save(scdata, config, task_name, sample_id, file = fpath_cont)
+        save(seurat_obj, config, task_name, sample_id, num_cells_to_downsample, file = fpath_cont)
```
seb-mueller, line 117: This change reminds me that we need to check for consistency of the filter steps in pipeline, i.e. sticking to either `seurat_obj` or `scdata`. Do you have any preference as to which? It's outside this scope but should be probably done by whoever next works on pipeline. Any

seb-mueller (approved):
lgtm

qc-runner/src/init.r
alexvpickering: I like `scdata` because it's less to type when im playing around :)

qc-runner/src/init.r
alexvpickering: Yeah me hardcoding it for saving is hacky .... not all the steps have a `num_cells_to_downsample` so we are going to get ugly somewhere

PR #66, opened 2021-05-24
fix per sample custom config

alexvpickering:
# Linked Issue

Also fixed in data-ingest: https://github.com/biomage-ltd/data-ingest/pull/34

# Background

The function `add_custom_config_per_sample` wasn't subsetting the dataset. This created a bug where the entire object was being used to calculate default cellSizeDistribution, numGenesVsNumUmis and doubletScore thresholds. For example all the doubletScore thresholds end up the same:

image

This is the same value as if I calculate the threshold with the full seurat object:

image

Whereas e.g. for an actual sample subset we get something different for each sample:

image

### DIFF
--- a/pipeline-runner/src/data-ingest/4_Prepare_experiment.r
+++ b/pipeline-runner/src/data-ingest/4_Prepare_experiment.r
@@ -104,7 +104,7 @@ add_custom_config_per_sample <- function(step_fn, config, scdata){
 
   samples <- seurat_obj$samples
   
-  for(sample in samples){
+  for(sample in unique(samples)){
     # Downsample the seurat object to a unisample experiment
     scdata_sample <- scdata[, samples %in% sample]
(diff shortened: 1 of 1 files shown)


alexvpickering:
It seems to have been a bug introduced by the fact that we have both `sample` and `samples` as meta.data columns:

```R
# samples was unique samples
samples <- unique(scseq$samples)
samples
# [1] "T0_1"   "T0_2"   "Ctrl_1" "Ctrl_2" "Gent_1" "Gent_2"

dim(scseq)
# [1]  18689 267867

dim(subset(scseq, samples %in% "T0_1"))
# [1] 18689  2103

sample <-  "T0_1"
dim(subset(scseq, samples %in% sample))
# [1]  18689 267867
```
I'm a fan of the slice notation because it's universal for R objects and clear when your subsetting columns/rows. Also not going to have this bug lol :)

PR #78, opened 2021-06-02
port doublet score code from data-ingest

alexvpickering:
# Background

scDblFinder is the current bottleneck in GEM2S. This should be faster and also prevent errors from asking for doublet scores when there are lots of empty droplets in the matrix.

Staging link:
https://ui-oliver-p78.scp-staging.biomage.net/
https://ui-mvpdev.scp-staging.biomage.net/


ogibson-biomage (commented):
Looks good. I'll stage and test on staging and approve once that's cleared

ogibson-biomage (approved):

PR #82, opened 2021-06-03
Fix default doublet config

alexvpickering:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&modal=detail&selectedIssue=BIOMAGE-1064

https://ui-alex-pipeline82.scp-staging.biomage.net/experiments/4223db357f54762ff657b0488004f7e8/data-processing

@ogibson-biomage This should also fix the issue we saw where there was both a `sample-` prefixed config and one without. The `sample-` prefixed config is not used anymore now that sample id's are hashes.

We should also do something so that the threshold has fewer digits but at least it is correct now

image

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


pipeline-runner/src/data-ingest/4_Prepare_experiment.r
```diff
@@ -107,7 +107,7 @@ add_custom_config_per_sample <- function(step_fn, config, scdata){
     result_config$auto <- NULL
     result_config$enabled <- NULL
     # Update config with the unisample thresholds
-    config[[paste("sample-", sample, sep = "")]] <- result_config
+    config[[sample]] <- result_config
```
ogibson-biomage, line 110: Is it correct to remove the sample- from the config? Might that break something?

pipeline-runner/src/doubletScores.r
```diff
@@ -7,14 +7,13 @@
...
 
-# The most uses values in doublet scores reporting in the scrublet paper [1] are around 0.25. There are not too much literature about how to compute
-# a threshold. For now, we will offer two methods:
-# --> Absolute threshold: In order to be not too extrictive the threshold is set to 0.25
-generate_default_values_doubletScores <- function(seurat_obj, config) {
-
-    # HARDCODE
-    threshold <- 0.25
-
+generate_default_values_doubletScores <- function(seurat_obj, sample) {
```
JuanluOnieva-biomage, line 10: I would suggest not changing the parameter config to sample, since this function should be the same of [this](https://github.com/biomage-ltd/pipeline/blob/89577460b5ef98babe095a509513bc2c15191d90/pipeline-runner/src/data-ingest/4_Prepare_experiment.r#L31-L39) (Moreover, in the data-ingest it seems that it was updated for scDblFinder, but in the pipeline it wasn't)

alexvpickering: The function you linked is being called [here](https://github.com/biomage-ltd/data-ingest/blob/0fbeb48c9eda8b9b1cf08e009ab3551a12af17ca/src/4_Prepare_experiment.r#L336) which first [subsets](https://github.com/biomage-ltd/pipeline/blob/89577460b5ef98babe095a509513bc2c15191d90/pipeline-runner/src/data-ingest/4_Prepare_experiment.r#L94) the suerat object so that only cells for a given sample are present.

In contrast here the seurat object is not a subset so we need the sample name in order to generate the appropriate per sample threshold when the user has selected _Automatic_. I don't beleive the `config` parameter was even used previously in this function.

pipeline-runner/src/data-ingest/4_Prepare_experiment.r
alexvpickering: It doesn't seem to. There may have been undiagnosed bugs without this change as the pipeline [grabs per-sample config](https://github.com/biomage-ltd/pipeline/blob/9af52b111ce1589994bf4c0b1d0304c94f6b66f5/pipeline-runner/src/init.r#L168) using the `sample_id` directly without the `sample-` prefix. Since it is not created currently, it is initually `NULL`. Only because `auto` is initially `TRUE`, the non `sample-` prefixed config is created during each step of the QC pipeline and then created after being [sent to the API](https://github.com/biomage-ltd/pipeline/blob/9af52b111ce1589994bf4c0b1d0304c94f6b66f5/pipeline-runner/src/handle_data.r#L20).

I tested locally by uploading Vicky's dataset without issue.

pipeline-runner/src/doubletScores.r
JuanluOnieva-biomage: Originally, we had a config function per step, where the structure was going to have `seurat_obj` and `config` as parameter of the functions. However, in the end it was not realized for all steps. So, I leave it up to you to filter inside the function or outside :)

I think the rest of the changes are all good, when we can test it in staging, I am happy to approve.

alexvpickering: staging link added. I uploaded with Vicky's dataset without obvious issues.

JuanluOnieva-biomage (approved):

pipeline-runner/src/data-ingest/4_Prepare_experiment.r
alexvpickering: @aerlaut - @JuanluOnieva-biomage mentioned that you sketched out the requirements when the `sample-` prefix was set up. There seems to have been a regression when data-ingest was moved over to the pipeline and the `sample-` prefix is not currently used (and hasn't been for a while now). Is there any critical reason for having a `sample-` prefix or if we just continue to use only the sample_id hash is that okay?

ogibson-biomage: Hey! I've actually realized today separately that this is a necessary change because it led to a weird accumulation of per samples config in the processingConfig and I raised a PR with this change, having forgotten that I saw the same thing in your PR. It'd still be good to have Agi's input on this and the plotdata but I think we should be good to go with this change.

alexvpickering: That is the rationale for the change. Juanlu pointed out though that we can also change the upload_to_aws.R step so that the initial sample's are named with `sample-` prefixes. This would also fix the accumulation of configs but be in line with @aerlaut original sketchup.

PR #92, opened 2021-06-07
Add tty flag to fix local output

xverges:
`-t` flag for docker run to fix the output and prevent annoying bells. 

https://github.com/apocas/dockerode#streams-goodness

image

The changes to `app.js` are only to cleanup the code.


alexvpickering (approved):
Tested locally. Much more pleasing to the eyes :)

PR #93, opened 2021-06-07
hardcode DEBUG_PATH

alexvpickering:
# Background
I updated my docker and it doesn't seem to like unset host part on volume anymore:

image

To resolve I've hardcoded the `DEBUG_PATH=$(pwd)/debug`. In addition to fixing this issue it also simplifies debug specification and always exposes a host path where it's possible to save something that will be available outside of the container. 

This provides a nice use patern that I find myself using: instead of specifying a DEBUG_STEP, you can just save something in `/debug`. I have used this for gem2s debugging where the DEBUG_STEP is not set up.

# Changes
### Code changes

# Definition of DONE

### Testing

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ make build
  access it later from `$(pwd)/debug`.
 
 To save the parameters (`config`, `seurat_obj`, etc) to a data processing task function, specify DEBUG_STEP.
-Available tasks include all those defined in `run_step` of  [init.r](pipeline-runner/src/init.r#L72) as well as `DEBUG_STEP=all` 
+Available tasks include all those defined in `run_processing_step` of  [init.r](pipeline-runner/src/init.r#L72) as well as `DEBUG_STEP=all` 
 to save the parameters to all data processing task functions:
 
 ```bash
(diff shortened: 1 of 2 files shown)


JuanluOnieva-biomage (commented):
Tested locally and it works fine for qc, however, if I try to debug a gm2s step, no `.RData` is saved; even if I use DEBUG_STEP=all only the qc steps are saved. Is it possible to debug gm2s? If it is not, we would have to update the readme, and maybe create a ticket for that.

ogibson-biomage (approved):

alexvpickering:
> Tested locally and it works fine for qc, however, if I try to debug a gm2s step, no `.RData` is saved; even if I use DEBUG_STEP=all only the qc steps are saved. Is it possible to debug gm2s? If it is not, we would have to update the readme, and maybe create a ticket for that.

adding gem2s steps will be a fair bit more involved. For now, the TLDR is you can save anything in /debug and it will be available on the host in $(pwd)/debug. This allows you to debug gem2s and requires less maintenance so I don't think it's worth adding a bunch of code at this point.

JuanluOnieva-biomage (approved):

PR #134, opened 2021-07-30
Refactor gem2s

alexvpickering:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1221

https://ui-apick-pipeline134.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


ogibson-biomage:
Hey Alex! You mentioned there was a bug on this one in scdblfinder. Is that fixed?

alexvpickering:
> Hey Alex! You mentioned there was a bug on this one in scdblfinder. Is that fixed?

I'll need to investigate more next week

pipeline-runner/R/qc-1-filter_emptydrops.R
```diff
@@ -98,7 +96,7 @@ filter_emptydrops <- function(scdata, config, sample_id, task_name = 'classifier
   # get filter stats after filtering
   filter_stats <- list(
     before = before,
```
ivababukova, line 98: take a look and see if there is a good reason to calculate before a bit differently than the other functions.

pipeline-runner/R/gem2s-7-upload_to_aws.R
```diff
@@ -1,97 +1,93 @@
...
+  # remove used colors from pool
+  used <- seq_along(samples_set$children)
+  color_pool <- color_pool[-used]
 
-  if (cluster_env == "production") {
-    print(sprintf("https://scp.biomage.net/experiments/%s/data-exploration", experiment_id))
+  # Design cell_set meta_data for DynamoDB
+  cell_sets <- c(list(scratchpad), list(samples_set))
+
+  # TODO: refactor meta_sets
```
ivababukova, line 80: take a look if this is not something that's already done

pipeline-runner/R/gem2s-4-score_doublets.R
```diff
@@ -1,65 +1,61 @@
...
+
+  scores <- list()
+  for (sample in samples) {
+    message("Sample --> ", sample, "...")
+    counts <- counts_list[[sample]]
+    edrops <- edrops_list[[sample]]
+
+    # scDblFinder expects empty droplets to be removed
+    if (!is.null(edrops)) {
+      keep <- which(edrops$FDR <= 0.001)
```
ivababukova, line 26: better make an R object or whatever it is called and add all those settings that we are using for gem2s, then get this 0.001 number from the object instead of hardcoding it

pipeline-runner/R/gem2s-3-run_emptydrops.R
```diff
@@ -1,64 +1,47 @@
...
-#' @param scdata Raw sparse matrix with the counts for one sample.
-#' @param sample_name Name of the sample that we are preparing.
-#'
-#' @return
-compute_emptydrops <- function(scdata, sample_name) {
-  message("Sample --> ", sample_name, "...")
+#' @param counts dgCMatrix with counts for one sample.
+compute_emptydrops <- function(counts) {
+  # check if filtered
+  nempty <- sum(Matrix::colSums(counts) < 100)
```
ivababukova, line 37: move to gem2s settings config and also the one on line 39

pipeline-runner/R/qc-1-filter_emptydrops.R
alexvpickering: changed to be like others

pipeline-runner/R/gem2s-7-upload_to_aws.R
alexvpickering: Hadn't done yet - nice catch!

pipeline-runner/R/gem2s-1-download_cellranger.R
```diff
@@ -1,29 +1,36 @@
...
   project_id <- input$projectId
   sample_names <- input$sampleNames
   sample_uuids <- input$sampleIds
 
-  print_config(1, "Cellranger", input, pipeline_config, list())
-
   s3 <- paws::s3(config = pipeline_config$aws_config)
 
-  fnames <- c("features.tsv.gz", "barcodes.tsv.gz", "matrix.mtx.gz")
+  cr_fnames <- c("features.tsv.gz", "barcodes.tsv.gz", "matrix.mtx.gz")
```
ivababukova, line 17: what does `cr_fnames` stand for?

pipeline-runner/R/gem2s-1-download_cellranger.R
```diff
@@ -47,13 +54,11 @@ download_cellranger <- function(input, pipeline_config) {
     config$metadata <- input$metadata
```
ivababukova, line 54: Lines 36 - 39: there is no error handling around the call to AWS. If the object is not in the S3 bucket, then this part of code will crash. Before we write the output on a file (line 42), we should check that download was ok, that the files are there. If they are not there, we should raise an exception.

pipeline-runner/R/gem2s-1-download_cellranger.R
```diff
@@ -47,13 +54,11 @@ download_cellranger <- function(input, pipeline_config) {
...
-
-  write(exportJSON, "/input/meta.json")
+  prev_out$config <- config
+  res <- list(
+    data = list(),
+    ouput = prev_out)
 
-  message("Step 1 complete")
-  return(list())
+  message("\nStep 1 complete")
```
ivababukova, line 62: Wouldn't it be more helpful if you said "Download of cellranger files step complete"? What if in the future we introduced a step before that one, then this will no longer be step 1.

pipeline-runner/R/gem2s-1-download_cellranger.R
```diff
@@ -1,29 +1,36 @@
...
+#' Download cellranger files from S3
+#'
+#' @param input The input object from the request
+#' @param pipeline_config result of \code{load_config}
+#' @param prev_out \code{list()} that is added to in each with gem2s task.
+#'
+#' @return list where 'output' slot has config used by \code{load_cellranger}
+#' @export
+#'
+download_cellranger <- function(input, pipeline_config, prev_out = list()) {
```
ivababukova, line 10: I would call this function `download_cellranger_files` instead

pipeline-runner/R/gem2s-1-download_cellranger.R
alexvpickering: cellranger filenames

pipeline-runner/R/gem2s-1-download_cellranger.R
alexvpickering: Let me know if this doesn't fully address the concern but in `init.R` function `init`, every task is wrapped in a `tryCatchLog` which, if anything fails will result in the pipeline returning an error message to the API and a callstack in the logs.

pipeline-runner/R/gem2s-1-download_cellranger.R
alexvpickering: I've gone ahead and changes to `cellranger_fnames`

pipeline-runner/R/gem2s-2-load_cellranger_files.R
```diff
@@ -0,0 +1,66 @@
...
+#'   \item{"counts_list"}{named list of dgCMatrix per sample}
+#'   \item{"annot"}{data.frame from reading features.tsv.gz for each sample}
+#' }
+#' @export
+#'
+load_cellranger_files <- function(input, pipeline_config, prev_out) {
+  message("Loading cellranger output ...")
+
+  # destructure previous output
+  config <- prev_out$config
```
ivababukova, line 16: what happens here if `config` is not in `prev_out`?

pipeline-runner/R/gem2s-1-download_cellranger_files.R
```diff
@@ -1,29 +1,36 @@
...
+    message("\nSample --> ", sample)
+
+    for (cellranger_fname in cellranger_fnames) {
+      gem_key <- file.path(project_id, sample, cellranger_fname)
 
-      message("GEM key")
-      message(gem_key)
+      message("GEM key: ", gem_key)
 
       sample_name <- sample_names[[match(sample, sample_uuids)]]
```
ivababukova, line 28: Is it possible to have `match` executed on `sample` that is not in `sample_uuids`? What will happen here in the code if this happens?

pipeline-runner/R/gem2s-1-download_cellranger_files.R
```diff
@@ -1,29 +1,36 @@
...
+  cellranger_fnames <- c("features.tsv.gz", "barcodes.tsv.gz", "matrix.mtx.gz")
   unlink("/input", recursive = TRUE)
 
   for (sample in sample_uuids) {
-    for (fname in fnames) {
-      gem_key <- file.path(project_id, sample, fname)
+    message("\nSample --> ", sample)
+
+    for (cellranger_fname in cellranger_fnames) {
+      gem_key <- file.path(project_id, sample, cellranger_fname)
```
ivababukova, line 24: Is it possible that one of the file names in `cellranger_fnames` will not be in `file.path`? What will happen in that case?

pipeline-runner/R/gem2s-2-load_cellranger_files.R
```diff
@@ -0,0 +1,66 @@
...
+    annot <- read.delim(annot_fpath, header = FALSE)
+
+    message(
+      sprintf("Sample %s has %s genes and %s droplets.", sample, nrow(counts), ncol(counts))
+    )
+
+    counts_list[[sample]] <- counts
+    annot_list[[sample]] <- annot
+  }
+  annot <- unique(do.call("rbind", annot_list))
```
ivababukova, line 61: what does calling `rbind` do?

pipeline-runner/R/gem2s-3-run_emptydrops.R
```diff
@@ -1,64 +1,47 @@
...
-    message("Found enough [", nr_barcodes_ambient, "] ambient barcodes,
-          so data is considered to be not filtered")
-    emptydrops_out <- DropletUtils::emptyDrops(scdata, lower = threshold_emptydrops)
-
-    file_path <- paste("/output/pre-emptydrops-", sample_name, ".rds", sep = "")
-    saveRDS(emptydrops_out, file = file_path, compress = FALSE)
+    out <- DropletUtils::emptyDrops(counts)
   }
+
+  return(out)
```
ivababukova, line 46: small thing, but I recommend small renamings to make it more obvious what this function does:
- rename `out`: give it a name to show that it is the empty droplets for a sample. For example, "sample_edrops".
- rename the function parameter `counts` to signify that these are per sample. For example, "sample_counts".
- rename function `compute_emptydrops` to `compute_sample_emptydrops` or even `compute_sample_edrops` so that it is consistent with the choice of `edrops` for empty droplets in all other places.

pipeline-runner/R/gem2s-3-run_emptydrops.R
```diff
@@ -1,64 +1,47 @@
...
-  # are confidently considered to be empty to serve as training set
-  nr_barcodes_ambient <- sum(colSums(scdata) < threshold_emptydrops)
-  range(colSums(scdata))
-  if (nr_barcodes_ambient < 50) {
-    flag_filtered <- TRUE
-    message("Found not enough [", nr_barcodes_ambient, "] ambient barcodes,
-          so data is considered to be pre-filtered")
-    message("Smallest nr of UMI for any barcode: ", min(colSums(scdata)))
-    message("EmptyDrops filter will not be applied")
+  if (nempty < gem2s$max.empty.drops) {
```
ivababukova, line 39: I am not sure I understand this if statement well. Is `nempty` equal to the amount of non-empty drops? If so, then why having less non-empty drops than empty ones means that the data is filtered? My thinking is that for the data to be considered as filtered, we need to have 0 empty drops. I am probably misunderstanding something, can you explain?

pipeline-runner/R/gem2s-4-score_doublets.R
```diff
@@ -1,68 +1,65 @@
...
+#' @return \code{prev_out} with added slot 'doublet_scores' containing result of call to
+#'   \code{\link{compute_doublet_scores}} for each sample.
+#'
+#' @export
+#'
+score_doublets <- function(input, pipeline_config, prev_out) {
+  message("Calculating probability of droplets being doublets...")
+
+  edrops_list <- prev_out$edrops
+  counts_list <- prev_out$counts_list
```
ivababukova, line 15: Are you validating and checking anywhere that these actually exist in `prev_out`? If step 3 was disabled or didn't filter anything, wouldn't these lines of code here break?

pipeline-runner/R/gem2s-4-score_doublets.R
```diff
@@ -1,68 +1,65 @@
...
+#'
+#' @inheritParams download_cellranger
+#' @param prev_out 'output' slot from call to \code{run_emptydrops}
+#'
+#' @return \code{prev_out} with added slot 'doublet_scores' containing result of call to
+#'   \code{\link{compute_doublet_scores}} for each sample.
+#'
+#' @export
+#'
+score_doublets <- function(input, pipeline_config, prev_out) {
```
ivababukova, line 11: again small recommendation: make the title of this function consistent with what you have for empty drops: call it `run_doublet_scores`

pipeline-runner/R/gem2s-4-score_doublets.R
```diff
@@ -1,68 +1,65 @@
...
-    keep <- which(edout$FDR <= 0.001)
-    scdata <- scdata[, keep]
-  }
-
-  # also filter low UMI as per scDblFinder:::.checkSCE()
-  ntot <- Matrix::colSums(scdata)
-  scdata <- scdata[, ntot > 200]
+#' @return data.frame with doublet scores and assigned classes
+#'
+compute_doublet_scores <- function(counts) {
```
ivababukova, line 53: rename this to `compute_sample_doublet_scores`, so that it is obvious that this is per sample, rename `counts` to `sample_counts` to make it obvious that this is for a sample.

pipeline-runner/R/gem2s-4-score_doublets.R
```diff
@@ -1,68 +1,65 @@
...
 
   set.seed(0)
-  scdata_DS <- scDblFinder(scdata, dbr = NULL, trajectoryMode = FALSE)
-  df_doublet_scores <- data.frame(
-    Barcodes = rownames(scdata_DS@colData),
-    doublet_scores = scdata_DS@colData$scDblFinder.score,
-    doublet_class = scdata_DS@colData$scDblFinder.class
+  sce <- scDblFinder::scDblFinder(counts)
+  dbl.df <- data.frame(
```
ivababukova, line 57: what is dbl.df?

pipeline-runner/R/gem2s-5-create_seurat.R
```diff
@@ -1,132 +1,129 @@
...
+#'
+create_seurat <- function(input, pipeline_config, prev_out) {
+
+  message("Creating Seurat Objects...")
+  message("items in prev_out: ", paste(names(prev_out), collapse = ' - '))
+
+  # destructure previous output
+  list2env(prev_out, envir = environment())
+
+  samples <- names(counts_list)
```
ivababukova, line 18: where did this `counts_list` come from? I don't see it in the function declaration, or anywhere else

pipeline-runner/R/gem2s-5-create_seurat.R
```diff
@@ -1,132 +1,129 @@
...
+  for (sample in samples) {
+    message("\nCreating SeuratObject for sample --> ", sample)
+
+    scdata_list[[sample]] <- construct_scdata(
+      counts = counts_list[[sample]],
+      doublet_score = doublet_scores[[sample]],
+      edout = edrops[[sample]],
+      sample = sample,
+      annot = annot,
+      config = config)
```
ivababukova, line 29: I am thinking here that all these came out of `prev_out` but it might be that my knowledge of R is not very great that I don't understand how exactly they were taken from there. Maybe the call of this function `list2env(prev_out, envir = environment())` on line 16 somehow magically loaded them in? @alexvpickering can you explain how they got here?

pipeline-runner/R/gem2s-5-create_seurat.R
```diff
@@ -1,132 +1,129 @@
...
 
-  message("Step 5 completed.")
-
-  return(list())
+  message("\nCreation of Seurat objects step complete.")
+  return(res)
 }
 
+# construct SeuratObject
+construct_scdata <- function(counts, doublet_score, edout, sample, annot, config, min.cells = 3, min.features = 10) {
```
ivababukova, line 43: better rename `edout` to `edrop_out`

ivababukova:
There is a file called `gem2s-5-test_object.R` that is put under folder `R`, it will be better to move it to be under the tests folder

pipeline-runner/R/gem2s-6-prepare_experiment.R
```diff
@@ -1,221 +1,174 @@
...
+#' @return prev_out \code{prev_out} with added slots 'scdata' containing merged
+#'   \code{SeuratObject} and 'qc_config' containing default config for QC steps.
+#'
+#' @export
+#'
+prepare_experiment <- function(input, pipeline_config, prev_out) {
+  message("Preparing experiment ...")
+
+  scdata_list <- prev_out$scdata_list
+  edrops <- prev_out$edrops
```
ivababukova, line 19: same question as on the previous files: can we actually count on these always to be in `prev_out`? What if they aren't and what can possibly cause them to not be there?

pipeline-runner/R/gem2s-6-prepare_experiment.R
```diff
@@ -1,221 +1,174 @@
...
+  # original name kept in 'original_name' column
+  gname <- annot$name
+  annot$original_name <- gname
   is.dup <- duplicated(gname) | duplicated(gname, fromLast = TRUE)
-#We need to convert the gene inputs from _ to - bc when we create the Seurat object we do this, and the match would return NA values if any of the inputs still has _. 
-  annotations$input <- gsub('_', '-', annotations$input)
-  annotations$name[is.dup] <- paste(gname[is.dup], annotations$input[is.dup], sep = " - ")
+
+  #We need to convert the gene inputs from _ to - bc when we create the Seurat object we do this, and the match would return NA values if any of the inputs still has _.
+  annot$input <- gsub('_', '-', annot$input)
```
ivababukova, line 39: why aren't we doing this in the step that is configuring the annotations and then putting then in `prev_out`?

pipeline-runner/R/gem2s-6-prepare_experiment.R
```diff
@@ -1,221 +1,174 @@
...
-  # In order to avoid duplicated genes names, we are going to add the ENSEMBL ID for those
-  # genes that are duplicated (geneNameDuplicated-ENSEMBL)
-  gname <- annotations$name
-  # Keep original name in 'original_name' variable
-  annotations$original_name <- gname
+  # add ENSEMBL ID for genes that are duplicated (geneNameDuplicated-ENSEMBL)
+  # original name kept in 'original_name' column
+  gname <- annot$name
+  annot$original_name <- gname
   is.dup <- duplicated(gname) | duplicated(gname, fromLast = TRUE)
```
ivababukova, line 36: why aren't we doing this in the step that is responsible for configuring the annotations?

pipeline-runner/R/gem2s-6-prepare_experiment.R
```diff
@@ -1,221 +1,174 @@
...
+#'  1) Merges the samples for the current experiment
+#'  2) Adds metadata: cellsId, color_pool, and gene annotation
+#'  3) Preparing QC configuration
+#'
+#' @inheritParams download_cellranger
+#' @param prev_out  'output' slot from call to \code{create_seurat}
+#'
+#' @return prev_out \code{prev_out} with added slots 'scdata' containing merged
+#'   \code{SeuratObject} and 'qc_config' containing default config for QC steps.
+#'
```
ivababukova, line 12: why don't you split each of those 3 items into separate functions? It will definitely make the code easier to read.

pipeline-runner/R/gem2s-6-prepare_experiment.R
ivababukova: Also, why preparing QC configuration here? It doesn't feel like this is the best place for constructing QC config, because it makes that step directly coupled to QC. It will be better to construct QC config just after calling the QC step, or at least, at the bare minimum move this code to a separate file, so that in the future if we want to add another step after "gem2s-prepare-experiment", we don't have to be copy-pasting and moving this QC code around.

ivababukova: I would probably move all this config to a file called `qc_config` or something similar. We can have a similar one for all the config that we have for gem2s too.

pipeline-runner/R/handle_data.R
```diff
@@ -153,17 +153,17 @@ upload_matrix_to_s3 <- function(pipeline_config, experiment_id, data) {
...
   message("Uploading updated count matrix to S3 bucket ", pipeline_config$processed_bucket, " at key ", object_key, "...")
-  s3$put_object(
-    Bucket = pipeline_config$processed_bucket,
-    Key = object_key,
-    Body = count_matrix
-  )
+  put_object_in_s3_multipart(
+    pipeline_config,
+    bucket = pipeline_config$processed_bucket,
+    object = count_matrix,
```
ivababukova, line 159: Agi did the same thing in one of his more recent PRs, I think it also got merged in. Will be best to pull master in to get these changes

pipeline-runner/data-raw/sysdata.R
```diff
@@ -0,0 +1,9 @@
+# constrants used in GEM2S
+gem2s <- list(
+    max.edrops.fdr = 0.001,
+    max.empty.counts = 100,
+    max.empty.drops = 50
+)
```
ivababukova, line 6: consider adding the cellranger_fnames from file `gem2s-1-download_cellranger_files`, line 17 because these are also configuration, as well as this `/input` dir name used in that file and elsewhere

pipeline-runner/data-raw/sysdata.R
```diff
@@ -0,0 +1,9 @@
...
+
+
+usethis::use_data(gem2s, internal = TRUE)
```
ivababukova, line 9: you can also make a similar file for that qc config that currently is added in step 6 of the gem2s pipeline.

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
```diff
@@ -0,0 +1,35 @@
+mock_counts <- function() {
```
ivababukova, line 1: an end-to-end test is missing that runs the whole gem2s pipeline with various inputs and various filters enabled and disabled.

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
```diff
@@ -0,0 +1,35 @@
...
+        'filtered'
+    )
+
+    expect_equal(out$output$edrops, list())
+})
+
+test_that("run_emptydrops runs when not pre-filtered", {
+    counts <- mock_counts()
+    prev_out <- list(counts_list = list(sample1=counts, sample2=counts))
+    out <- run_emptydrops(NULL, NULL, prev_out)$output
```
ivababukova, line 24: what about a test for running emptydrops when filter 2 or filter 1 or both have been skipped previously?

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
```diff
@@ -0,0 +1,35 @@
+mock_counts <- function() {
+    set.seed(0)
+    DropletUtils:::simCounts()
+}
+
+
+test_that("run_emptydrops is skipped when dataset is pre-filtered", {
```
ivababukova, line 7: what else can possibly fail in the empty droplets filter that we can test for?

pipeline-runner/R/gem2s-1-download_cellranger_files.R
alexvpickering: Not possible as `sample` is from interating through `sample_uuids`

pipeline-runner/R/gem2s-2-load_cellranger_files.R
alexvpickering: `call_read10x` will fail

pipeline-runner/R/gem2s-2-load_cellranger_files.R
alexvpickering: `do.call(rbind, list)` takes a list of `data.frame`s and stacks them by rows. the `unique` then makes sure there aren't any duplicate rows. The purpose is to have a data.frame of all the unique feature annotations.

pipeline-runner/R/gem2s-3-run_emptydrops.R
alexvpickering: `nempty` is the number of obviously empty droplets which is defined as the number of columns with fewer than a threshold number of counts (100). Typically the vast majority of droplets are empty (100k++). If there are fewer than `max.empty.drops` (50), we consider the dataset filtered.

pipeline-runner/R/gem2s-3-run_emptydrops.R
alexvpickering: Good suggestion - done :)

pipeline-runner/R/gem2s-2-load_cellranger_files.R
ivababukova: so let's add a guard here to check if `config` is there before we try to access it, if it is not there, then raise an error ourselves.

pipeline-runner/R/gem2s-1-download_cellranger_files.R
ivababukova: I see

pipeline-runner/R/gem2s-2-load_cellranger_files.R
ivababukova: I understand, thanks for explaining!

pipeline-runner/R/gem2s-1-download_cellranger_files.R
alexvpickering: the `file.path` is used to construct the S3 object location that we are going to download. If they are not there then the upload failed earlier on but gem2s was allowed to start anyways. If that happened then an error will be thrown when it tries to get the S3 object from S3. I 'm not sure that checking that the S3 objects are there and throwing an error if they are not provides additional benefit. If they are not there, GEM2S can't run and we have to throw an error anyways. It would just add code/time to list the S3 objects first with the same endpoint - throwing an error.

pipeline-runner/R/gem2s-4-score_doublets.R
alexvpickering: If edrops isn't there it will be fine (`edrops_list` will be `NULL`) (in fact we expect it to not be there if it is a filtered dataset).  If `counts_list` isn't there it will fail as it should because it means we either don't have any count matrices or we didn't pass them on and there is no way for GEM2S to complete. I'll add a check for it.

pipeline-runner/R/gem2s-4-score_doublets.R
alexvpickering: The GEM2S function names don't follow a convention that they begin with `run_`. For example we have `load_cellranger_files`, `create_seurat`, `prepare_experiment`, `upload_to_aws`.

pipeline-runner/R/gem2s-4-score_doublets.R
alexvpickering: a `data.frame` with results from `sc"Dbl"finder`. I made it more explicit

pipeline-runner/R/gem2s-5-create_seurat.R
alexvpickering: Its from the `list2env` call. There were several things to take from `prev_out` so this simplifies destructuring but I can make it all more explicit

pipeline-runner/R/gem2s-5-create_seurat.R
alexvpickering: Yes that is correct. It takes a list of named objects and add them to the environment specified. It's a more succinct version of this:

```R
  # destructure previous output
  config <- prev_out$config
  annot <- prev_out$annot
  edrops <- prev_out$edrops
  counts_list <- prev_out$counts_list
  doublet_scores <- prev_out$doublet_scores
```

pipeline-runner/R/gem2s-5-create_seurat.R
alexvpickering: > There is a file called `gem2s-5-test_object.R` that is put under folder `R`, it will be better to move it to be under the tests folder

Yes @ogibson-biomage created it as a test of the `SeuratObject` before completing gem2s but then disabled it. Is it okay to remove @ogibson-biomage or are we going to restore it's functionality?

pipeline-runner/R/gem2s-6-prepare_experiment.R
alexvpickering: I think that these sorts of tests are better done for the functions that generate and store them (e.g. testing that the function `create_seurat` adds a item to `prev_out` named `scdata_list`.

 `edrops` can be absent and will be fine. if `scdata_list` wasn't created then the function will fail because we don't have `SeuratObjects`

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
alexvpickering: GEM2S doesn't have the concept of skipping/disabling previous steps. All steps of GEM2s run every time.

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
alexvpickering: GEM2S steps can't be disabled (unlike QC steps). They are all run every time. I agree that we should add end-to-end test of gem2s pipeline but I think that this should be part of a seperate ticket as this one has already grown quite large.

pipeline-runner/data-raw/sysdata.R
alexvpickering: step 6 of gem2s is creating qc config that is uploaded to AWS DynamoDB. This is a set of constants that are used by gem2s. I agree that we should do the same for QC constants although I think that could be part of a larger refactor that we want to undertake for QC pipeline as well and is probably best done as a seperate ticket to keep the scope of this ticket on GEM2S

pipeline-runner/R/gem2s-6-prepare_experiment.R
alexvpickering: Not sure - was there any reason to but this here @ogibson-biomage or can I move it to `load_cellranger` ?

pipeline-runner/R/gem2s-6-prepare_experiment.R
alexvpickering: We put it here because QC config is the initial default QC config that QC will be run with.  It gets uploaded to dynamodb and forms the request parameters when QC is initially triggered. We don't have it after calling the QC step because the QC steps don't know if they are running QC for the first time (in which case the parameters are the defaults created here) or the user has changed parameters.

alexvpickering: I'll move the QC config to another file

pipeline-runner/R/gem2s-6-prepare_experiment.R
alexvpickering: Will do

pipeline-runner/R/handle_data.R
alexvpickering: Will do

alexvpickering:
> There is a file called `gem2s-5-test_object.R` that is put under folder `R`, it will be better to move it to be under the tests folder

Yes @ogibson-biomage created it as a test of the SeuratObject to be run during every run of gem2s but then disabled it. Is it okay to remove Oliver or are we going to restore it's functionality?

pipeline-runner/R/gem2s-3-run_emptydrops.R
ivababukova: Thanks for explaining. So `nempty` stands for "number of initially empty droplets", right? Btw, where did that `threshold_emptydrops` variable from line 42 of the old changes go?

ivababukova: I would consider renaming `nempty` to something that makes it more obvious that this is a number of empty droplets like: "default_empty_count", or "numb_empty" or something like that

pipeline-runner/R/gem2s-4-score_doublets.R
ivababukova: I would consider introducing a little convention, because it can be confusing which function is being the one called and which is the one internal to the file. 
I would also recommend defining the ones that are only internally used as private, but I don't think you can have private methods in R without having classes. Or can you?

pipeline-runner/R/gem2s-5-create_seurat.R
ivababukova: I think it is fine as it is now with the comment on top. Just again, make sure you are checking that everything that is compulsory is in prev_out

pipeline-runner/data-raw/sysdata.R
ivababukova: sure, happy to do it as part of another ticket, this PR is definitely getting way too big

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
ivababukova: of course, I forgot that we are not able to disable steps, good point :) An end to end test as well as tests for the individual steps will definitely be super useful.

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
ivababukova: yes, of course, forgot that

pipeline-runner/R/gem2s-1-download_cellranger.R
ivababukova: It will definitely error out at some point because of that, but looking at the code, it looks like it will error out much later after the error occurs, which will make it harder to debug. So I think what would happen if there is no such object in the bucket is:

1. Lines 36 to 39 will run successfully. There will be no error raised there. body will be either empty or undefined.
2. If body is evaluated to None/null, then line 42 will most likely run successfully. writeBin function will just write an empty content to the supplied `local_fpath`.
3. The next place where we try to read from `local_fpath` and do things on the contents of `local_fpath` will fail,  because it will be empty. This `tryCatchLog` check in `init.R` will detect it and raise an error at the place where we are trying to do things with the empty contents of `local_fpath`. 

The problem with this flow is that the error in 3. won't show that the real problem comes from 1. This will take us much more time to trace back and debug the error. This is why a good practice is to try and detect and raise issues as early as possible. 
So in this case with this code, after lines 36 to 39 I would add a check if body is empty, and if it is, I will raise an error right there.

pipeline-runner/R/gem2s-6-prepare_experiment.R
ogibson-biomage: Keep in mind that in construct_scdata (per sample) we are using annot

ogibson-biomage: But I'm not sure that is a problem

pipeline-runner/R/qc-1-filter_emptydrops.R
```diff
@@ -97,14 +99,14 @@ filter_emptydrops <- function(scdata, config, sample_id, task_name = 'classifier
 
   # get filter stats after filtering
   filter_stats <- list(
-    before = before,
-    after = calc_filter_stats(scdata, tmp_sample)
+    before = calc_filter_stats(scdata, sample_id),
```
ogibson-biomage, line 102: Like this changes

pipeline-runner/renv.lock
```diff
@@ -559,13 +538,6 @@
       "Repository": "CRAN",
       "Hash": "a0cbe758a531d054b537d16dff4d58a1"
     },
-    "doParallel": {
```
ogibson-biomage, line 562: All of these are no longer used? Why where they in the original snapshot?

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
ogibson-biomage: our gem2s "end-to-end" test is the test_seurat_object function, which checks that the object has the correct slots. It doesn't however cross-check that the data is valid.

ogibson-biomage (commented):
Can't approve right now because the staging link is failing: steps get to 100% and stop

pipeline-runner/R/gem2s-6-prepare_experiment.R
ivababukova: to my non-bioinformatician eye it looks like:
- in this file, we are substituting `_` with `-`. This means that if we move this code to be executed before construct_scdata, then line 85 in `add_mito`: `mt.features <- annot$input[grep("^mt-", annot$name, ignore.case = TRUE)]` would cause everything that was originally with `^mt_` to be classified as `mt.features` too. Not too sure if it is a problem or not.
It is not a big deal if we have to do this annot changes in this file, I suggested it because I thought it will be more tidy there

pipeline-runner/tests/testthat/test-gem2s-3-run_emptydrops.R
ivababukova: why don't we move it to be where the other tests are?

pipeline-runner/R/gem2s-4-score_doublets.R
ivababukova: I thought it would be a pain, this is why I didn't really suggest it. But ideally when you have a couple of methods, some of them are only intended for internal use, it is good to restrict them so that in the future they don't get misused and also so that the code is more readable. It is not a huge deal though

pipeline-runner/init.R
```diff
@@ -195,6 +199,7 @@ call_data_processing <- function(task_name, input, pipeline_config) {
     # upload plot data result to S3
     plot_data_keys <- send_plot_data_to_s3(pipeline_config, experiment_id, rest_of_results)
     tstart <- Sys.time()
+    plot_data_keys <- send_plot_data_to_s3(pipeline_config, experiment_id, rest_of_results)
```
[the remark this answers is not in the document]
alexvpickering: good catch. Will remove

pipeline-runner/R/gem2s-1-download_cellranger_files.R
alexvpickering: I just checked and if the object isn't in S3 then the error is thrown right away (I changed the gem key to download blah.tsv.gz):

image

pipeline-runner/R/gem2s-4-score_doublets.R
alexvpickering: For GEM2S all the main functions that are also endpoints are the only functions that are exported. All other GEM2S utilities/helpers don't have the `@export` tag so if the R code is installed as a package they are internal. 

As Oliver mentioned though we do not install the package when we run it in a docker container but instead source all the R code, making everything available. The upside of doing it this way is that the watcher works (the pipeline hot reloads).

ivababukova: I see, sounds reasonable

ogibson-biomage (approved):

PR #147, opened 2021-09-02
init visium pipeline

alexvpickering:
Hackathon yo!


pipeline-runner/R/vis2s-1-process_spaceranger_files.R
```diff
@@ -0,0 +1,31 @@
+process_spaceranger_files <- function(input, pipeline_config, prev_out) {
+    scdata_list <- list()
```
ogibson-biomage, line 2: We need to sort out the gene annotations! And cell ids generation

PR #165, opened 2021-09-23
relax code coverage

alexvpickering:
With current settings, a PR that reduces the number of lines will fail. For example (3 lines less, 3 fewer hits, so fail?): 

Screenshot from 2021-09-22 19-14-04

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -2,16 +2,16 @@ codecov:
   require_ci_to_pass: yes
 
 coverage:
-  range: 20..100
+  range: 40..100
   round: down
   precision: 2
 
   status:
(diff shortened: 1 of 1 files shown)


kafkasl:
@alexvpickering we have intentionally changed this. we are very concerned that a lot of small issues find their way into production because our tests are sorely lacking. Although having the codecov fail just because of removing lines is not ideal, in that situation we should all take the chance to write some more tests. In this case the coverage is 36% so quite far from ideal.

ivababukova:
Completely agree with Pol here. The quality and quantity of our tests is a concern, so having a negative code coverage report is a perfect opportunity to go back to the code and add a few tests for what is left untested.

PR #170, opened 2021-09-27
use HOST_IP for api url

alexvpickering:
# Background
`host.docker.internal` doesn't exist for linux. This uses `HOST_IP` for linux as set in the makefile.

Staging link

https://ui-olivergibson-pipeline170.scp-staging.biomage.net/


ogibson-biomage (approved):
Works locally, need to test on staging

ogibson-biomage (approved):

PR #188, opened 2021-11-23
fix renv setup

alexvpickering:
# Background
While debugging [another issue](https://biomage.atlassian.net/jira/software/c/projects/BIOMAGE/boards/1?modal=detail&selectedIssue=BIOMAGE-1556) I noticed this after `exec`'ing into the pipeline container and starting an interactive R session. It seems that updates in `renv.lock` were not necessarily taking effect. I encountered this in another project I worked on and so ported over the same solution

Before change:
Screenshot from 2021-11-22 13-48-52

After change:

Screenshot from 2021-11-22 16-35-00

This didn't resolve the above issue but needed to be fixed as well.


alexvpickering:
closing as checked in production and doesn't seem to be an issue

PR #212, opened 2022-03-07
[BIOMAGE-1755] - Upload debug to s3

alexvpickering:
# Description
uploads dump file and logs to S3 when pipeline throws an error

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1755

https://ui-alex-pipeline212.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates
- [x] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [x] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [x] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [x] Unit tests written **or** no unit tests required for change, e.g. documentation update.

### Integration testing

- [x] Started end-to-end tests on the latest commit.

### Documentation updates
- [x] Relevant Github READMEs updated **or** no GitHub README updates required.
- [x] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional


pipeline-runner/R/handle_data.R
```diff
@@ -250,6 +250,19 @@ upload_matrix_to_s3 <- function(pipeline_config, experiment_id, data) {
   return(object_key)
 }
 
+upload_debug_folder_to_s3 <- function(debug_subdir, pipeline_config) {
+  fnames <- list.files(file.path("/debug", debug_subdir))
```
ivababukova, line 254: this `/debug` folder name is used all over the code. How about you define a config file (like we have in the python and in the r worker) and in then you move it there in the config folder as a config variable and then import and read it in various places in the code?
Btw, we will need to do this for multi-account deployments for aws resource names and aws region and so on

pipeline-runner/R/handle_data.R
```diff
@@ -250,6 +250,19 @@ upload_matrix_to_s3 <- function(pipeline_config, experiment_id, data) {
...
+
+  message("Uploading logs and dump file to S3 bucket ", pipeline_config$debug_bucket, " with prefix ", debug_subdir, "...")
```
ivababukova, line 256: `debug_subdir` is actually experiment id + the timestamp. I see that you generate the timestamp in `init.R`, but the  timestamp actually belongs in this function and it is used only here. I suggest that:
- you remove the timestamp generation from `init.R` and add it inside this function
- remove this `debug_subdir` in `init.R` and only pass in `input$experimentId` to `upload_debug_folder_to_s3`
- in `handle_data.R`, rename the parameter called `debug_subdir` to `experimentId`.
- in `handle_data.R` generate the debug_subdir (but better call it `s3_path`) by concatenating `experimentId` with the calculated timestamp, adding a `/` between them

pipeline-runner/R/handle_data.R
ivababukova: I actually see now that you already have config defined `init.R`. So I would add the debug folder in there too

pipeline-runner/init.R
```diff
@@ -337,6 +339,14 @@ init <- function() {
             quit('no')
         }
 
+        # parse data from state machine input
+        input <- RJSONIO::fromJSON(input_json, simplify = FALSE)
+
+        # save logs to file
+        debug_subdir <- file.path(input$experimentId, debug_timestamp)
+        debug_path <- file.path("/debug", debug_subdir)
+        flog.appender(appender.tee(file.path(debug_path, "logs.txt")))
```
ivababukova, line 348: what's this `flog.appender` doing?

pipeline-runner/R/handle_data.R
```diff
@@ -250,6 +250,19 @@ upload_matrix_to_s3 <- function(pipeline_config, experiment_id, data) {
   return(object_key)
 }
 
+upload_debug_folder_to_s3 <- function(debug_subdir, pipeline_config) {
```
ivababukova, line 253: can you add unit tests for this function?

pipeline-runner/R/handle_data.R
alexvpickering: yes this is something we are aware of and should be addressed. The same is true of `work.R` in the worker. They are both largely untested

pipeline-runner/R/handle_data.R
ivababukova: I see. In this case, I would move the logic of generating the s3 prefix string inside the `upload_debug_folder_to_s3 ` function. To the `upload_debug_folder_to_s3`, I will just pass the timestamp and the experiment id

alexvpickering: The benefit is just that the prefix is already generated in init.R as it is used there seperately. I will duplicate it inside `upload_debug_folder_to_s3` if you would like

pipeline-runner/init.R
alexvpickering: that line specifies a file to write logs to in addition to the console

pipeline-runner/R/handle_data.R
ivababukova: I didn't realise that it's used in init.R. When I looked, I didn't see the usage. In this case, better to keep the generation only in init.R

ogibson-biomage (approved):
I think this looks good. Not many comments to add, Iva has already commented a lot about specific code. I checked and we already have expiration rules for data in the bucket which was one of my main concerns.

I haven't personally tested on staging, I think that's the final thing before merging.

gerbeldo (approved):
tested in staging, LGTM!

PR #233, opened 2022-05-23
Check empty features

saracastel:
# Description
Checks if there are any features with empty gene symbol both in the features file and in the count matrix, and remove them if < 0.1% of the total number of features

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1907

#### Link to staging deployment URL (or set N/A)
https://ui-sara-pipeline233.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist
Your changes will be ready for merging after all of the steps below have been completed.

### Code updates
Have best practices and ongoing refactors being observed in this PR
- [ ] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [x] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [ ] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [ ] Unit tests written **or** no unit tests required for change, e.g. documentation update.

### Integration testing
**You must check the box below** to run integration tests on the latest commit on your PR branch.
Integration tests have to pass before the PR can be merged. Without checking the box, your PR
**will not pass** the required status checks for merging.

- [ ] Started end-to-end tests on the latest commit.

### Documentation updates
- [ ] Relevant Github READMEs updated **or** no GitHub README updates required.
- [ ] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional
- [ ] Staging environment is unstaged before merging.
- [ ] Photo of a cute animal attached to this PR.

PR #241, opened 2022-06-01
[BIOMAGE-1817] Atypical features files

ogibson:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1817

https://ui-olivergibson-p241.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates
- [ ] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [X] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [x] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [x] Unit tests written **or** no unit tests required for change, e.g. documentation update.

### Integration testing

- [ ] Started end-to-end tests on the latest commit.

### Documentation updates
- [ ] Relevant Github READMEs updated **or** no GitHub README updates required.
- [ ] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional


ogibson:
Untitled Diagram drawio

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
   rownames(annot) <- annot$input
   return(annot)
 }
+
+# Fix annotations makes annotations compatible between samples with different types.
+# The possible options for features_types at this stage are 0,1,2
+#  0 is SYMBOL/ SYMBOL
+#  1 is IDS/SYMBOL
+#  2 is IDS/IDS
+equalize_annotation_types <- function(annot_list, counts_list, features_types_list, samples) {
```
gerbeldo, line 263: Why use numbers and not strings? "sym-sym", "id-sym", "id-id" would be clearer.

gerbeldo: also, roxygen2 docs!

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+  if (any(features_types_list == 2) && any(features_types_list == 0) && !any(features_types_list == 1)) stop("Incompatible features detected.")
+
```
gerbeldo, line 265:
```suggestion
  if (any(features_types_list == 2) &&
      any(features_types_list == 0) &&
      !any(features_types_list == 1)) {
      stop("Incompatible features detected.")
}

```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+  if (any(features_types_list == 1) && (any(features_types_list == 2) || any(features_types_list == 0))) {
```
gerbeldo, line 266:
```suggestion
  if (any(features_types_list == 1) &&
     (any(features_types_list == 2) || any(features_types_list == 0))) {

```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+
+    for (sample in samples) {
+      if (features_types_list[[sample]] == 0 || features_types_list[[sample]] == 2) {
+        sample_annot <- annot_list[[sample]]
+
+        # Try to replace input column (currently symbols) in sample_annot with ids from annots_with_ids
+        if (features_types_list[[sample]] == 0) {
+
+          matched_symbols_index <- match(sample_annot$input,annots_with_ids$name)
+          is_in_annot_list <- which(sample_annot$input %in% annots_with_ids$name)
```
gerbeldo, line 279:
```suggestion
          matched_symbols_index <- match(sample_annot$input, annots_with_ids$name)
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+    }
+  }
+  return(list(counts_list=counts_list, annot_list=annot_list))
+}
+
+#' extract_features_types
+#'
+#' Determines the type of an annot data frame.
+#' Classifies the columns into either ensemblIds or symbols, and extracts a number that represents the combination.
+#'
```
gerbeldo, line 310:
```suggestion
#' Classifies the columns into either ensemblIds or symbols, and extracts a
#' number that represents the combination.
```

pipeline-runner/NAMESPACE
```diff
@@ -2,6 +2,7 @@
 
 export(create_seurat)
 export(download_user_files)
+export(extract_features_types)
```
gerbeldo, line 5: I would change features_types to feature_types. Let's not get a new cells_id 🙏

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+    annots_with_ids <- annots_with_ids[!duplicated(annots_with_ids$input), ]
+
+    for (sample in samples) {
+      if (features_types_list[[sample]] == 0 || features_types_list[[sample]] == 2) {
+        sample_annot <- annot_list[[sample]]
+
+        # Try to replace input column (currently symbols) in sample_annot with ids from annots_with_ids
+        if (features_types_list[[sample]] == 0) {
+
+          matched_symbols_index <- match(sample_annot$input,annots_with_ids$name)
```
gerbeldo, line 278:
```suggestion
          matched_symbols_index <- match(sample_annot$input, annots_with_ids$name)
```

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: I went with numbers so they can be generated by adding bools, but feel free to change it!

pipeline-runner/NAMESPACE
ogibson: sounds good!

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -54,14 +54,15 @@ load_user_files <- function(input, pipeline_config, prev_out, input_dir = "/inpu
...
+  features_types_list <- list()
 
   samples <- config$samples
 
-
   for (sample in samples) {
     sample_dir <- file.path(input_dir, sample)
     sample_fpaths <- list.files(sample_dir)
     annot_fpath <- file.path(sample_dir, "features.tsv.gz")
+    gene_column <- 1
```
kafkasl, line 65: ideally add constants either at the top of the file or a different one and use upper case (or any convention you are following) to make it more readable

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -54,14 +54,15 @@ load_user_files <- function(input, pipeline_config, prev_out, input_dir = "/inpu
...
   annot_list <- list()
+  features_types_list <- list()
 
   samples <- config$samples
 
-
   for (sample in samples) {
     sample_dir <- file.path(input_dir, sample)
     sample_fpaths <- list.files(sample_dir)
     annot_fpath <- file.path(sample_dir, "features.tsv.gz")
```
kafkasl, line 64: 1. if this `features.tsv.gz` is a constant I'd add it to the top of the file or something.
2. how does this work if the file is called `genes.tsv.gz` which we say we support in the UI?

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -71,19 +72,38 @@ read_10x_files <- function(config, input_dir) {
...
 
-    counts <- Seurat::Read10X(sample_dir, gene.column = 1)
+    annot <- read.delim(annot_fpath, header = FALSE)
 
-    if (is(counts, "list")) {
-      slot <- "Gene Expression"
-      # questionable: grab first slot if no gene expression
-      if (!slot %in% names(counts)) slot <- names(counts)[1]
-      counts <- counts[[slot]]
+    if (ncol(annot) == 1 || annot[1, 2] == "Gene Expression") {
```
kafkasl, line 77: All of this code dealing with the annotations is very hard to read to me or any engineer without the actual seurat object. First I'd suggest to move all inside the loop intro 3 different functions: one for producing the `annotations` and the other to produce the `counts_list`. `e.g.`:
```
    annot <- extract_annotations(annot_fpath)
    features_types <- extract_feature_types(annot)
    counts <- extract_counts(sample_dir, annot, feature_types)
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -71,19 +72,38 @@ read_10x_files <- function(config, input_dir) {
...
+      annot[, 2] <- annot[, 1]
     }
 
-    annot <- read.delim(annot_fpath, header = FALSE)
+    features_types <- extract_features_types(annot)
+
+    message("Features types is ", features_types, "for sample ", sample)
+
+    if (features_types == -1) {
```
kafkasl, line 85: if `-1` has a meaning like `not_found` considering naming it as a constant, or adding a comment

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -71,19 +72,38 @@ read_10x_files <- function(config, input_dir) {
...
+      annot[, c(1, 2)] <- annot[, c(2, 1)]
+      gene_column <- 2
+      features_types <- 1
+    }
```
kafkasl, line 89: the previous if is imopssible to read for anyone without the seurat object, moreover I think it's very error prone. I'd try to make it more readable by either adding an explanation or naming the hardcoded values as constants (if it makes sense)

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
   rownames(annot) <- annot$input
   return(annot)
 }
+
+# Fix annotations makes annotations compatible between samples with different types.
```
kafkasl, line 258: Fix annotations does not seem to be the name of the function anymore; also `equalize` sounds a bit weird, maybe it's me but the name did not make it clear what the function does.

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+# The possible options for features_types at this stage are 0,1,2
+#  0 is SYMBOL/ SYMBOL
+#  1 is IDS/SYMBOL
+#  2 is IDS/IDS
```
kafkasl, line 262: what are the differences between `ids` and `symbol` ?

pipeline-runner/R/gem2s-2-load_user_files.R
kafkasl: where could it be useful to generate them as bools? I'd probably go also with strings

pipeline-runner/R/gem2s-2-load_user_files.R
kafkasl: all the hardcoded values should be constants. e.g.:
```
if (any(features_types_list == IDS_IDS) &&
	      any(features_types_list == SYM_IDS) &&
	      !any(features_types_list == SYM_SYM)) {
```

pipeline-runner/R/gem2s-2-load_user_files.R
kafkasl: same with constants

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+equalize_annotation_types <- function(annot_list, counts_list, features_types_list, samples) {
+  if (any(features_types_list == 2) && any(features_types_list == 0) && !any(features_types_list == 1)) stop("Incompatible features detected.")
+
+  if (any(features_types_list == 1) && (any(features_types_list == 2) || any(features_types_list == 0))) {
+    annots_with_ids <- unique(do.call("rbind", annot_list[features_types_list == 1]))
+
+    annots_with_ids <- annots_with_ids[!duplicated(annots_with_ids$input), ]
+
+    for (sample in samples) {
+      if (features_types_list[[sample]] == 0 || features_types_list[[sample]] == 2) {
```
kafkasl, line 272: add hardcoded values as constants

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+          #The counts have been loaded with gene symbols so we need to replace the rownames with the ids
+          rownames(counts_list[[sample]]) <- sample_annot$input
+        }
+
+        # Try to replace names column (currently ids) in sample_annot with symbols from annots_with_ids
+        if (features_types_list[[sample]] == 2) {
+          matched_ids_index <- match(sample_annot$input,annots_with_ids$input)
+          is_in_annot_list <- which(sample_annot$input %in% annots_with_ids$input)
+
+          sample_annot$name[is_in_annot_list] <- annots_with_ids$name[na.omit(matched_ids_index)]
```
kafkasl, line 296: so in here you are not replacing all the names of the sample, just some of them, what happens to the other ones that are not in `is_in_annot_list` ?

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+#  2 is IDS/IDS
+equalize_annotation_types <- function(annot_list, counts_list, features_types_list, samples) {
+  if (any(features_types_list == 2) && any(features_types_list == 0) && !any(features_types_list == 1)) stop("Incompatible features detected.")
+
+  if (any(features_types_list == 1) && (any(features_types_list == 2) || any(features_types_list == 0))) {
+    annots_with_ids <- unique(do.call("rbind", annot_list[features_types_list == 1]))
+
+    annots_with_ids <- annots_with_ids[!duplicated(annots_with_ids$input), ]
+
+    for (sample in samples) {
```
kafkasl, line 271: this for looks complex but it might be clearer as:
```
for (sample in samples) {

    if (features_types_list[[sample]] == SYM_SYM) { 

        sample_annot$input <- input_sym_to_ids(annot_list[[sample]])   
        rownames(counts_list[[sample]]) <- sample_annot$input

    } else if (features_types_list[[sample]] == IDS_IDS) {
         sample_annot$name <- name_ids_to_sym(sample_annot)
    }
}
```
so you can read the loop easily and then add a short comment on functions `input_sym_to_ids` and `name_ids_to_sym` to explain what they do

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+
+  annot_c1 <- annot[, 1]
+  is_ens <- annot_c1[substr(annot_c1, 1, 3) == "ENS"]
+  features_types[[1]] <- length(is_ens) >= length(annot_c1) - length(is_ens)
+
+  annot_c2 <- annot[, 2]
+  is_ens <- annot_c2[substr(annot_c2, 1, 3) == "ENS"]
+  features_types[[2]] <- length(is_ens) >= length(annot_c2) - length(is_ens)
+
+  if (features_types[[1]] == FALSE && features_types[[2]] == TRUE) {
```
kafkasl, line 332: I don't get this `feature_types` have been one of those SYM_SYM, IDS_IDS, etc... values but here it's suddenly a boolean? it looks like it's a boolean resulting from comparing `lenghts` so naming it `types` does not make sense.

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: They are already renamed by this stage I believe, we would need to check the UI but I'm fairly certain that this was discussed when we originally wrote this

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: This is a variable though (gene_column)

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: I like this idea. Let's add this to sysdata.R

pipeline-runner/R/gem2s-2-load_user_files.R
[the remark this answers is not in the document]
ogibson: In this case it's not an index, it's a true/false boolean specifying if it's in annot_list or not

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+      if (features_types_list[[sample]] == 0 || features_types_list[[sample]] == 2) {
+        sample_annot <- annot_list[[sample]]
+
+        # Try to replace input column (currently symbols) in sample_annot with ids from annots_with_ids
+        if (features_types_list[[sample]] == 0) {
+
+          matched_symbols_index <- match(sample_annot$input,annots_with_ids$name)
+          is_in_annot_list <- which(sample_annot$input %in% annots_with_ids$name)
+
+          sample_annot$input[is_in_annot_list] <- annots_with_ids$input[na.omit(matched_symbols_index)]
```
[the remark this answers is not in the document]
ogibson: because all the non matching ids will be returned as NA instead of positions and the indexing in annots_with_ids$input will fail if those NAs are present

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: I made this to make this function easier to write instead of ending in three:

if(a) return a
if(b) return b
if(c) return c

ogibson: Maybe that option will be clearer though

ogibson: But basically in this function we are still working with booleans and from the combination of bools we return an int that represents the current state of the features file

ogibson: I do love the idea of replacing all numbers for constant numbers as in 

IDS_IDS
SYM_SYM

Etc.

ogibson: I was going to add a comment to change it to a dictionary but constants sound much better

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: They will remain with the same name. There are no annotations for those, we can't replace the names with something else

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: We do need to do this

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: You are missing an important if at the beginning, and the important operation of replacing 

annot_list[[sample]] <- sample_annot

ogibson: That being said, I think we can do this for all samples regarding of type, the result will be that for most we will just replace the same thing, and that's what I wanted to avoid. 

It's a matter of clarity vs an extra assignation, but I think in this case the assignation won't be costly so we could potentially replace the for with:

for (sample in samples) { 
     sample_annot <- annot_list[[sample]]
     if (features_types_list[[sample]] == SYM_SYM) { 
          sample_annot$input <- input_sym_to_ids(annot_list[[sample]]) 
          rownames(counts_list[[sample]]) <- sample_annot$input 
 } else if (features_types_list[[sample]] == IDS_IDS) { 
          sample_annot$name <- name_ids_to_sym(sample_annot) 
 } 
      annot_list[[sample]] <- sample_annot
 }

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: Gene IDS are unique identifiers with species information. Gene symbols are a mess, and usually different species might have the same symbol for a different gene id

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: Yeah also sounded weird to me, but it's trying to make all the annotations equal. I'll think alternatives

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: This is complex because there's basically no alternative. It's a table with two columns and we need to reverse columns 1 and 2

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: -1 is actually

SYM_IDS

So when we replace with constants it will be ok

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: Keep in mind that at this stage there's no seurat object, we are just working with a table with 2-3 columns

ogibson: I originally wanted to refactor this for into functions, and I almost pushed that change but eventually decided that it's simpler to read in this way if you are following the flow of what's going on with the tables. @gerbeldo what do you think?

We can certainly do this if both of you agree that it will look better.

pipeline-runner/R/gem2s-2-load_user_files.R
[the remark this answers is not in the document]
alexvpickering: `which` returns an index:

```R
which(letters[1:5] %in% 'd')
[1] 4
```

pipeline-runner/R/gem2s-2-load_user_files.R
alexvpickering: why the need for ints? Just make the constants strings. That way when we debug and print `feature_types` we get `IDS_IDS` instead of `0` or `-1`.

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+#' @return
+#' -1 is SYMBOl/IDS
+#' 0 is SYMBOL/ SYMBOL
+#' 1 is IDS/SYMBOL
+#' 2 is IDS/IDS
+#'
+#'
+#' @examples
+extract_features_types <- function(annot) {
+  features_types <- list()
```
alexvpickering, line 322: is there a need to have this as a list vs a vector? Just makes indexing less verbose down the line:

```R
features_types[1] vs features_types[[1]]
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+
+  annot_c1 <- annot[, 1]
+  is_ens <- annot_c1[substr(annot_c1, 1, 3) == "ENS"]
+  features_types[[1]] <- length(is_ens) >= length(annot_c1) - length(is_ens)
+
+  annot_c2 <- annot[, 2]
+  is_ens <- annot_c2[substr(annot_c2, 1, 3) == "ENS"]
+  features_types[[2]] <- length(is_ens) >= length(annot_c2) - length(is_ens)
```
alexvpickering, line 330: more than half of values in column 2 start with string ENS?

```R
is_ens_col1<- startsWith(annot[[1]], "ENS")
pct_ens_col1 <- sum(is_ens_col1) / nrow(annot)

is_ens_col2 <- startsWith(annot[[2]], "ENS")
pct_ens_col2 <- sum(is_ens_col2) / nrow(annot)

if (pct_ens_col1 > 0.5 & pct_ens_col2 > 0.5) return(IDS_IDS)
```

pipeline-runner/R/gem2s-2-load_user_files.R
alexvpickering, line 332: more than half of values in column start with string ENS?

```R
is_ens_col1 <- startsWith(annot[[1]], "ENS")
pct_ens_col1 <- sum(is_ens_col1) / nrow(annot)

is_ens_col2 <- startsWith(annot[[2]], "ENS")
pct_ens_col2 <- sum(is_ens_col2) / nrow(annot)

if (pct_ens_col1 > 0.5 & pct_ens_col2 > 0.5) return(IDS_IDS)
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -234,3 +254,84 @@ format_annot <- function(annot_list) {
...
+#'
+#' @return
+#' -1 is SYMBOl/IDS
+#' 0 is SYMBOL/ SYMBOL
+#' 1 is IDS/SYMBOL
+#' 2 is IDS/IDS
+#'
+#'
+#' @examples
+extract_features_types <- function(annot) {
```
alexvpickering, line 321:
```suggestion
detect_feature_types <- function(annot) {
```

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: they are renamed in the UI, and in S3 always stored as features.tsv.gz

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: I am refactoring a bit, given the conflicts with #233.

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: what about normalise or standardise?

gerbeldo: or normalize/standardize if you'd rather go _'murica style_

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: I like this. stolen 👍

gerbeldo: replaced all ints with character variables defined in sysdata.R

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: done

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: changed the hardcoded ints to strings defined in sysdata.R

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: done

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +263,134 @@ format_annot <- function(annot_list) {
...
+  is_ens <- c(pct_ens_col1 >= 0.5, pct_ens_col2 >= 0.5)
+
+  # reverse case, sym in first and id in second column
+  if (!is_ens[1] && is_ens[2]) return(SYM_IDS)
+
+  # regular cases. sum of booleans returns ints. convert to char to string match
+  feature_type <- switch(as.character(sum(is_ens)),
+                         "0" = SYM_SYM,
+                         "1" = IDS_SYM,
+                         "2" = IDS_IDS)
```
gerbeldo, line 354: I know this is ugly, I'm open to suggestions that don't involve 3 ifs

gerbeldo: given how `switch` works, the other alternative, also ugly, would be:

```suggestion
  # reverse case, sym in first and id in second column
  if (!is_ens[1] && is_ens[2]) return(SYM_IDS)

  # regular cases. sum of booleans returns ints. convert to char to string match
  feature_type <- switch(sum(is_ens) + 1,
                         SYM_SYM,
                         IDS_SYM,
                         IDS_IDS)
```

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: Also adding some comments to the obscure parts, such as this.

pipeline-runner/R/gem2s-2-load_user_files.R
gerbeldo: hey @ogibson, don't we prefer count matrices with ENS ID's as rownames? I understand that from the fact that gene_column defaults to 1. So why do you change it here?

gerbeldo: oooh, I see. Read10X does not look to the annot object, but reads it again from the original features file

pipeline-runner/R/gem2s-2-load_user_files.R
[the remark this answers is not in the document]
gerbeldo: renamed to symbols_with_ids

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,38 @@ parse_rhapsody_matrix <- function(config, input_dir) {
...
 }
 
+read_10x_annotations <- function(annot_fpath, sample) {
+  gene_column <- 1
+
+  annot <- read.delim(annot_fpath, header = FALSE)
+
+  # duplicate first col if there is only one col with gene names/ids, or
+  # if there is a "Gene Expression" col
+  if (ncol(annot) == 1 || annot[1, 2] == "Gene Expression") {
```
kafkasl, line 219: `annot[1, 2] == "Gene Expression"` I'm having a hard time with this in R. The indexing `annot[1, 2]` means return the first and second column? how do you return first and second row?. In any case, it will return two values right? but you can directly compare it to a string? Is it going to make an element-wise comparison and then if any of those is true, return tru?

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,38 @@ parse_rhapsody_matrix <- function(config, input_dir) {
...
+    annot[, 2] <- annot[, 1]
+  }
+
+  feature_types <- get_feature_types(annot)
+
+  message("Feature types are ", feature_types, "for sample ", sample)
+
+  # reverse annot cols if symbols are first
+  if (feature_types == SYM_IDS) {
```
kafkasl, line 228: I assume that having the gene IDS first, and the symbols second is a convention

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,38 @@ parse_rhapsody_matrix <- function(config, input_dir) {
   return(list(counts_list = counts_list, annot = annot))
 }
 
+read_10x_annotations <- function(annot_fpath, sample) {
```
kafkasl, line 212: while I more or less understand the steps inside this function, I'm having a hard time understanding the overall goal of the function. Maybe a short description would be helpful. Specially about what's the final desired format of the annotations.

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,38 @@ parse_rhapsody_matrix <- function(config, input_dir) {
...
+  # reverse annot cols if symbols are first
+  if (feature_types == SYM_IDS) {
+    annot[, c(1, 2)] <- annot[, c(2, 1)]
+    gene_column <- 2
+    feature_types <- IDS_SYM
+  }
+
+  # make same names in annot seurat makes for count matrix.
+  # Only first col, we either copy or reverse order of the annot data.frame
+  annot[, 1] <- make.unique(annot[, 1])
```
kafkasl, line 236: if we make unique the column, how does it match the other column later?

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,38 @@ parse_rhapsody_matrix <- function(config, input_dir) {
   return(list(counts_list = counts_list, annot = annot))
 }
 
+read_10x_annotations <- function(annot_fpath, sample) {
+  gene_column <- 1
+
+  annot <- read.delim(annot_fpath, header = FALSE)
+
+  # duplicate first col if there is only one col with gene names/ids, or
+  # if there is a "Gene Expression" col
```
gerbeldo, line 218: change comment to reflect the fact that some features files are missing columns

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +265,225 @@ format_annot <- function(annot_list) {
...
+#'
+#' @export
+get_feature_types <- function(annot) {
+  is_ens_col1 <- startsWith(annot[[1]], "ENS")
+  pct_ens_col1 <- sum(is_ens_col1) / nrow(annot)
+
+  is_ens_col2 <- startsWith(annot[[2]], "ENS")
+  pct_ens_col2 <- sum(is_ens_col2) / nrow(annot)
+
+  is_ens <- c(pct_ens_col1 >= 0.5, pct_ens_col2 >= 0.5)
```
gerbeldo, line 344: change variable to is_ids

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,38 @@ parse_rhapsody_matrix <- function(config, input_dir) {
   return(list(counts_list = counts_list, annot = annot))
 }
 
+read_10x_annotations <- function(annot_fpath, sample) {
+  gene_column <- 1
```
gerbeldo, line 213: maybe change to "seurat_rownames_col" or add comment explaining

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -71,66 +71,36 @@ read_10x_files <- function(config, input_dir) {
...
+    annotations <- read_10x_annotations(annot_fpath, sample)
+    counts <- Seurat::Read10X(sample_dir, gene.column = annotations[["gene_column"]], unique.features = TRUE)
 
     if (is(counts, "list")) {
       slot <- "Gene Expression"
-      # questionable: grab first slot if no gene expression
+      # questionable: grab first slot if no slot named gene expression
       if (!slot %in% names(counts)) slot <- names(counts)[1]
       counts <- counts[[slot]]
     }
```
gerbeldo, line 82: add comment about discarding elements in list. why we discard the rest of the matrices

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +265,225 @@ format_annot <- function(annot_list) {
...
+#' @export
+#'
+sym_to_ids <- function(sample_annot, annot_with_ids) {
+  matched_symbols_index <-
+    match(sample_annot$input, annot_with_ids$name)
+  symbols_with_ids <-
+    which(sample_annot$input %in% annot_with_ids$name)
+
+  sample_annot$input[symbols_with_ids] <-
+    annot_with_ids$input[na.omit(matched_symbols_index)]
```
gerbeldo, line 400: we need @ogibson to explain this

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: annot[1,2] will return row 1, column 2 of the data frame. In this case we want to check if the second column is filled of elements named "Gene Expression" because that would mean that we have a column of gene ids/gene symbols and a column indicating the type of measurement. 

If we have less than 2 columns, that means we only have the feature symbols/ids, so it's safe to duplicate the first column as well. 

If we have more than one column or the second column is different from gene expression, that means we have two columns with ids/symbols and a third column with type of measurement.

All these cases are based on experience and cases we have actually seen in real life datasets.

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: Why don't you like this? I think it's a fine alternative to all the ifs, and is more readable than the sum of bools that I originally wrote

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson: Named this way because the read10x parameter has this name

pipeline-runner/R/gem2s-2-load_user_files.R
ogibson, line 218:
```suggestion
  # Some feature files have less columns than expected.
  # Duplicate first column if there is only one column with gene names/ids, or
  # if there is a "Gene Expression" column
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +266,225 @@ format_annot <- function(annot_list) {
...
+#'
+#' @export
+get_feature_types <- function(annot) {
+  is_ens_col1 <- startsWith(annot[[1]], "ENS")
+  pct_ens_col1 <- sum(is_ens_col1) / nrow(annot)
+
+  is_ens_col2 <- startsWith(annot[[2]], "ENS")
+  pct_ens_col2 <- sum(is_ens_col2) / nrow(annot)
+
+  is_ens <- c(pct_ens_col1 >= 0.5, pct_ens_col2 >= 0.5)
```
ogibson, line 345:
```suggestion
  is_ens <- c(pct_ens_col1, pct_ens_col2) >= 0.5
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,40 @@ parse_rhapsody_matrix <- function(config, input_dir) {
...
+  # reverse annot cols if symbols are first
+  if (feature_types == SYM_IDS) {
+    annot[, c(1, 2)] <- annot[, c(2, 1)]
+    # Read10x reads from the feature file which is not modified, but contains IDS in the second column. Use gene_column to indicate to Read10x from where to read the ids. 
+    gene_column <- 2
+    feature_types <- IDS_SYM
+  }
+
+  # make same names in annot seurat makes for count matrix.
+  # Only first col, we either copy or reverse order of the annot data.frame
```
ogibson, line 237:
```suggestion
  # Make the names in annot the same as the ones in the Read10x generated count matrix
  # Since Seurat uses makes.unique, we need to as well.
  # Only for the first column, at this stage column 1 should be IDs.
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,41 @@ parse_rhapsody_matrix <- function(config, input_dir) {
...
+  if (feature_types == SYM_IDS) {
+    annot[, c(1, 2)] <- annot[, c(2, 1)]
+    # Read10x reads from the feature file which is not modified, but contains IDS in the second column. Use gene_column to indicate to Read10x from where to read the ids. 
+    gene_column <- 2
+    feature_types <- IDS_SYM
+  }
+
+  # Make the names in annot the same as the ones in the Read10x generated count matrix
+  # Since Seurat uses makes.unique, we need to as well.
+  # Only for the first column, at this stage column 1 should be IDs.
```
ogibson, line 238:
```suggestion
  # Only for the first column, at this stage column 1 are the matrix rownames.
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -239,6 +209,41 @@ parse_rhapsody_matrix <- function(config, input_dir) {
...
+  if (feature_types == SYM_IDS) {
+    annot[, c(1, 2)] <- annot[, c(2, 1)]
+    # Read10x reads from the feature file which is not modified, but contains IDS in the second column. Use gene_column to indicate to Read10x from where to read the ids. 
+    gene_column <- 2
+    feature_types <- IDS_SYM
+  }
+
+  # Make the names in annot the same as the ones in the Read10x generated count matrix
+  # Since Seurat uses makes.unique, we need to as well.
+  # Only for the first column, at this stage column 1 are the matrix rownames.
```
ogibson, line 238:
```suggestion
  # Only for the first column, at this stage column 1 are the counts matrix rownames.
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,225 @@ format_annot <- function(annot_list) {
...
+
+    rownames(counts)[unnamed_genes_idx][keep_ids] <-
+      annotations$annot$name[unnamed_genes_idx][keep_ids]
+
+    rownames(annotations$annot)[unnamed_genes_idx][keep_ids] <-
+      annotations$annot$name[unnamed_genes_idx][keep_ids]
+
+    annotations$annot$input[unnamed_genes_idx][keep_ids] <-
+      annotations$annot$name[unnamed_genes_idx][keep_ids]
+
```
ogibson, line 474:
```suggestion
    available_gene_symbols <- annotations$annot$name[unnamed_genes_idx][keep_ids]
    
    rownames(counts)[unnamed_genes_idx][keep_ids] <- available_gene_symbols

    rownames(annotations$annot)[unnamed_genes_idx][keep_ids] <- available_gene_symbols

    annotations$annot$input[unnamed_genes_idx][keep_ids] <- available_gene_symbols

```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+    message("Replaced ", length(which(keep_ids)),
+            "empty gene names with available not empty annotation.")
+
+  }
+
+  # remove remaining rows if there are any
+  if (any(!keep_ids)) {
+  counts <- counts[-unnamed_genes_idx[!keep_ids], ]
+  annotations$annot <- annotations$annot[-unnamed_genes_idx[!keep_ids], ]
+
```
ogibson, line 482:
```suggestion
  genes_to_remove <- unnamed_genes_idx[!keep_ids]
  counts <- counts[-genes_to_remove, ]
  annotations$annot <- annotations$annot[-genes_to_remove, ]
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+    annotations$annot$input[unnamed_genes_idx][keep_ids] <- available_gene_symbols
+
+    message("Replaced ", length(which(keep_ids)),
+            "empty gene names with available not empty annotation.")
+
+  }
+
+  # remove remaining rows if there are any
+  if (any(!keep_ids)) {
```
ogibson, line 479: add indentation

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+#'
+#' @param sample_annot data.frame of annotations
+#' @param annot_with_ids data.frame of annotations with IDs. Key data.frame
+#'
+#' @return
+#' @export
+#'
+sym_to_ids <- function(sample_annot, annot_with_ids) {
+  matched_symbols_index <-
+    match(sample_annot$input, annot_with_ids$name)
```
ogibson, line 398:
```suggestion
 symbols_index_in_annot <-
    match(sample_annot$input, annot_with_ids$name)
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+  symbols_with_ids <-
+    which(sample_annot$input %in% annot_with_ids$name)
+
+  sample_annot$input[symbols_with_ids] <-
+    annot_with_ids$input[na.omit(matched_symbols_index)]
```
ogibson, line 403:
```suggestion
    annot_with_ids$input[na.omit(symbols_index_in_annot)]
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+#' existing symbol.
+#'
+#' @param sample_annot data.frame of annotations
+#' @param annot_with_ids data.frame of annotations with IDs. Key data.frame
+#'
+#' @return
+#' @export
+#'
+sym_to_ids <- function(sample_annot, annot_with_ids) {
+ symbols_index_in_annot <-
```
ogibson, line 397:
```suggestion
  symbols_index_in_annot <-
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+  symbols_with_ids <-
+    which(sample_annot$input %in% annot_with_ids$name)
+
+  sample_annot$input[symbols_with_ids] <-
+    annot_with_ids$input[na.omit(symbols_index_in_annot)]
+
+  #This avoids duplicates after combining with the annotated df.
+  #Leads to a mismatch in genes between samples but it seems like the best solution
+  sample_annot$input <- make.unique(sample_annot$input)
+
```
ogibson, line 408:
```suggestion
  symbol_idx_in_annot <-
    match(sample_annot$input, annot_with_ids$name)
  symbols_with_ids <-
    which(sample_annot$input %in% annot_with_ids$name)

  sample_annot$input[symbols_with_ids] <-
    annot_with_ids$input[na.omit(symbol_idx_in_annot)]

  #This avoids duplicates after combining with the annotated df.
  #Leads to a mismatch in genes between samples but it seems like the best solution
  sample_annot$input <- make.unique(sample_annot$input)

```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+  symbols_with_ids <-
+    which(sample_annot$input %in% annot_with_ids$name)
+
+  sample_annot$input[symbols_with_ids] <-
+    annot_with_ids$input[na.omit(symbol_idx_in_annot)]
+
+  #This avoids duplicates after combining with the annotated df.
+  #Leads to a mismatch in genes between samples but it seems like the best solution
+  sample_annot$input <- make.unique(sample_annot$input)
+
```
ogibson, line 408:
```suggestion
  symbols_with_idx_in_annot <-
    which(sample_annot$input %in% annot_with_ids$name)

  sample_annot$input[symbols_with_ids] <-
    annot_with_ids$input[na.omit(symbols_with_idx_in_annot)]

  #This avoids duplicates after combining with the annotated df.
  #Leads to a mismatch in genes between samples but it seems like the best solution
  sample_annot$input <- make.unique(sample_annot$input)

```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+#' @return
+#' @export
+#'
+sym_to_ids <- function(sample_annot, annot_with_ids) {
+  symbol_idx_in_annot <-
+    match(sample_annot$input, annot_with_ids$name)
+  symbols_with_idx_in_annot <-
+    which(sample_annot$input %in% annot_with_ids$name)
+
+  sample_annot$input[symbols_with_ids] <-
```
ogibson, line 402:
```suggestion
  sample_annot$input[symbols_with_idx_in_annot] <-
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+    which(sample_annot$input %in% annot_with_ids$name)
+
+  sample_annot$input[symbols_with_idx_in_annot] <-
+    annot_with_ids$input[na.omit(symbols_with_idx_in_annot)]
+
+  #This avoids duplicates after combining with the annotated df.
+  #Leads to a mismatch in genes between samples but it seems like the best solution
+  sample_annot$input <- make.unique(sample_annot$input)
+
+  return(sample_annot)
```
ogibson, line 409:
```suggestion
  symbol_idx_in_annot <-
    match(sample_annot$input, annot_with_ids$name)
  symbols_with_ids_in_annot <-
    which(sample_annot$input %in% annot_with_ids$name)

  sample_annot$input[symbols_with_ids_in_annot] <-
    annot_with_ids$input[na.omit(symbols_with_ids_in_annot)]

  #This avoids duplicates after combining with the annotated df.
  #Leads to a mismatch in genes between samples but it seems like the best solution
  sample_annot$input <- make.unique(sample_annot$input)

  return(sample_annot)
```

pipeline-runner/R/gem2s-2-load_user_files.R
```diff
@@ -263,3 +268,223 @@ format_annot <- function(annot_list) {
...
+#'
+ids_to_sym <- function(sample_annot, annot_with_ids) {
+  matched_ids_index <- match(sample_annot$input, annot_with_ids$input)
+  symbols_with_ids <-
+    which(sample_annot$input %in% annot_with_ids$input)
+
+  sample_annot$name[symbols_with_ids] <-
+    annot_with_ids$name[na.omit(matched_ids_index)]
+
+  return(sample_annot)
```
ogibson, line 432:
```suggestion
  ids_index_in_annot <- match(sample_annot$input, annot_with_ids$input)
  ids_with_symbols_in_annot <-
    which(sample_annot$input %in% annot_with_ids$input)

  sample_annot$name[ids_with_symbols_in_annot] <-
    annot_with_ids$name[na.omit(ids_index_in_annot)]

  return(sample_annot)
```

pipeline-runner/tests/testthat/test-gem2s-2-load_user_files.R
```diff
@@ -285,213 +231,284 @@ test_that("load_user_files loads 10x multisample experiments", {
...
-  expect_true("counts_list" %in% names(res))
-  expect_true(names(samples) %in% names(res$counts_list))
-  expect_s4_class(res$counts_list[[1]], "dgCMatrix")
+  res <- pipeline:::normalize_annotation_types(annot_list,counts_list,features_types_list, samples)
+
+  expect_equal(res[[1]],counts_list)
+  expect_equal(res[[2]],annot_list)
 })
 
+test_that("equalize_annotations does nothing if there are no samples with annotations",{
```
ogibson, line 307:
```suggestion
test_that("normalize_annotation_types does nothing if there are no samples with annotations",{
```

kafkasl (approved):

PR #315, opened 2023-08-08
add automated cluster detection for Seurat object uploads

alexvpickering:
# Description

# Details
#### URL to issue
N/A

N/A

#### Integration test branch
master

# Merge checklist

### Code updates
- [ ] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [ ] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [ ] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [ ] Unit tests written **or** no unit tests required for change, e.g. documentation update.

### Integration testing

- [ ] Started end-to-end tests on the latest commit.

### Documentation updates
- [ ] Relevant Github READMEs updated **or** no GitHub README updates required.
- [ ] Relevant Wiki pages created/updated **or** no Wiki updates required.

### Optional


pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
```diff
@@ -64,6 +77,69 @@ upload_seurat_to_aws <- function(input, pipeline_config, prev_out) {
...
+    for (exclude_col in exclude_cols)  {
+      exclude_vals <- meta[[exclude_col]]
+      if (test_groups_equal(check_vals, exclude_vals)) {
+        is_sample_col <- TRUE
+        break
+      }
+    }
+    if (is_sample_col) next()
+
+    # TODO: decide if should skip if values are boolean?
```
gerbeldo, line 123: I'm curious about this. what would be the usecase?

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
```diff
@@ -96,7 +172,7 @@ add_metadata_to_input <- function(scdata, input) {
 }
 
 # get column names that are consistent with sample groups
-find_group_columns <- function(scdata) {
+find_group_columns <- function(scdata, remove.dups = TRUE) {
```
gerbeldo, line 175: unrelated to the PR, but collapse interfaces nicely with dplyr, fndistinct could be used here as well.

pipeline-runner/tests/testthat/test-seurat-2-load_seurat.R
```diff
@@ -195,9 +195,45 @@ test_that("load_seurat fails if there if HVFInfo can't be called", {
...
+  scdata <- res$output$scdata
+  disp_colnames <- c('mean', 'variance', 'variance.standardized', 'ENSEMBL', 'SYMBOL')
+  expect_true(methods::is(scdata@misc$gene_dispersion, 'data.frame'))
+  expect_equal(colnames(scdata@misc$gene_dispersion), disp_colnames)
+
+  # clean up
+  unlink(data_dir, recursive = TRUE)
+})
+
+test_that("load_seurat identifies and log transforms counts stored in logcounts assay", {
```
gerbeldo, line 210: isn't RNA assay data slot instead of logcounts assay?

gerbeldo (commented):
small comments on the code, but looks good. still need to stage and test.

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
alexvpickering: Just that a column could have boolean values e.g. `isDoublet` or `isEpithelial`. We don't necessarily want to include the former as a cellset but want to include the later. It's really just a particular way of encoding a two-valued cellset division (which would pass all other checks) so I elected not to exclude them for the initial implementation.

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
```diff
@@ -64,6 +77,69 @@ upload_seurat_to_aws <- function(input, pipeline_config, prev_out) {
...
+  meta <- meta |> dplyr::relocate(dplyr::any_of(louvain_cols))
+
+  check_cols <- setdiff(colnames(meta), exclude_cols)
+
+  cluster_cols <- c()
+  for (check_col in check_cols) {
+    check_vals <- meta[[check_col]]
+
+    # skip if too few or way too many values
+    n.vals <- length(table(check_vals))
```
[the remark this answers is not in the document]
alexvpickering: Thanks! I restructured it a bit as I was calculating the same thing multiple times

old:
```
98: n.vals <- length(table(check_vals))
...
108: value_counts <- table(check_vals)
...
110: length(unique(check_vals))
```

I elected not to use `collapse::fndistict` as it is the `table(check_vals)` that is the slower part but is still necessary and now only being called once. Probably a bit of premature optimization though (~0.02 seconds for your example with half a million cells) :wink:.

pipeline-runner/R/seurat-2-load_seurat.R
```diff
@@ -53,53 +53,75 @@ reconstruct_seurat <- function(dataset_fpath) {
...
 
-  # add dispersions and gene annotations to new scdata
-  scdata@misc$gene_dispersion <- dispersions
+  # add logcounts
+  tryCatch({
+    logcounts <- user_scdata[['RNA']]@data
+    test_user_sparse_mat(logcounts)
+
+    # shouldn't be raw counts
+    suspect.counts <- max(logcounts) > 100
```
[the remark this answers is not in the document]
alexvpickering: Totally agree. My motivation here wasn't about speed but rather that I have had a user with a Seurat object where the `@counts` slot was replaced with logcounts. I didn't want to explicitly block the user from exploring the logcounts and warned them that they should not trust any of the between group differential expression analyses. If I were to test for `@data` and `@counts` being identical for this user, it would be `TRUE` and the logcounts would be double-logged.

I think the heuristic is fairly safe as it would take a count value of `x = 2^100 = 1.3e+30` to get `log2(x) = 100`.

 The Seurat object itself was from the [FlyCellAtlas](https://scope.aertslab.org/#/FlyCellAtlas/FlyCellAtlas%2Fr_fca_biohub_body_10x.loom/gene)
 
 We should probably also, at minimum, warn (or potentially disallow) between group DE if the same condition is detected for Seurat projects.

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
gerbeldo: yeah, one-hot encoding features makes all of this more complicated. It's something that could be required to address by the users (I'm also considering this case for the cell-level metadata feature)

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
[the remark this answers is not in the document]
gerbeldo: totally makes sense! it's a recent discovery and I'm excited 😬 . Some functionality is even faster than data.table!

alexvpickering:
> small comments on the code, but looks good. still need to stage and test.

Any idea why the Description is being overwritten? I'm guessing the above is from the description in `biomage-org`

pipeline-runner/R/seurat-2-load_seurat.R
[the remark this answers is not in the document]
gerbeldo: I see. Makes sense. I agree on the warning or disallowing between groups DE.  I suspect trajectory analysis would also be affected by the lack of raw counts.

pipeline-runner/R/seurat-2-load_seurat.R
```diff
@@ -53,53 +53,75 @@ reconstruct_seurat <- function(dataset_fpath) {
...
+    message(e$message)
+    stop(errors$ERROR_SEURAT_HVFINFO, call. = FALSE)
+  })
+
+  # add gene annotations
+  gene_annotations <- data.frame(
+    input = rns,
+    name = rns,
+    original_name = rns,
+    row.names = rns
```
gerbeldo, line 123: you could check if the gene_annotations exist in the seurat object, and just use that (testing for df safety). Or are gene names handled elsewhere?

This would add support for rds objects downloaded from Cellenics (some extra step might be required downstream though, don't remember now).

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
```diff
@@ -15,15 +15,28 @@ upload_seurat_to_aws <- function(input, pipeline_config, prev_out) {
...
+  if (!length(cluster_columns))
+    stop(errors$ERROR_SEURAT_CLUSTERS, call. = FALSE)
+
+  # add clusters
+  cluster_sets <- list()
+  for (i in seq_along(cluster_columns)) {
+    col <- cluster_columns[i]
+
+    # first cluster column used as louvain clusters
+    method <- ifelse(i == 1, 'louvain', col)
```
gerbeldo, line 29: the double louvain issue I reported stems from here. There was a "louvain" column in the metadata, but it was not the first one in the metadata table.

pipeline-runner/R/seurat-2-load_seurat.R
```diff
@@ -53,53 +53,75 @@ reconstruct_seurat <- function(dataset_fpath) {
...
+  error = function(e) {
+    message(e$message)
+    stop(errors$ERROR_SEURAT_LOGCOUNTS, call. = FALSE)
+  })
+
+
+  # add dispersions (need for gene list)
+  tryCatch({
+    # don't use first 3 columns
+    dispersions <- Seurat::FindVariableFeatures(logcounts)[-3]
```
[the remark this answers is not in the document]
alexvpickering: Thanks for the catch! The comment was incorrect. I changed the code to make it explicit about what columns it is keeping

pipeline-runner/R/seurat-2-load_seurat.R
alexvpickering: I don't really see that as the use-case of Seurat object upload. If the project was processed in Cellenics, you should look at the original project so that you can have access to all of the processing options and associated information in Data Processing.

pipeline-runner/R/seurat-3-upload_seurat_to_aws.R
alexvpickering: Yes, [this fixes it](https://github.com/hms-dbmi-cellenics/pipeline/pull/315/files#diff-6fe272316d0d7d696b8e23088d8e97aa9d24dee988ecf6ea74daac7921244d56R88)

pipeline-runner/R/seurat-2-load_seurat.R
gerbeldo: given this, I'd say it's good to go!

alexvpickering: I added it -- was relatively simple and I think it makes sense from the perspective that it simplifies things for developers (can use Seurat objects downloaded from Cellenics for testing). Thanks for the suggestion!

